PR #2833, opened 2023-07-24
Fix python version discovery

### LINKED ISSUE #2832: Fix python discovery
Cmake doesnt discover the currently activated environment if there is a higher python version available

je-cook:
Closes #2832 
This does a couple of small things, the main one is set the cmake policy to CMP0094 (https://cmake.org/cmake/help/latest/policy/CMP0094.html#policy:CMP0094). 

This set a few new defaults the first is `SET(Python3_FIND_VIRTUALENV FIRST)` which was being used anyway.

Secondly it sets `SET(Python3_FIND_STRATEGY LOCATION)` (see https://cmake.org/cmake/help/latest/module/FindPython3.html#hints) which is the fix to find the first python in your path (eg the activated environment) if it fulfills all the other conditions (eg min version).

I have changed the minimum cmake version which was was actually already required for `SET(Python3_FIND_VIRTUALENV FIRST)`

I've tested locally while fiddling with getting process updated for bluemira.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,21 +1,22 @@
 # CMake file for Pythonic PROCESS
-# Author  :   K. Zarebski (UKAEA)
-# Date    :   last modified 2020-11-09
+# Author  :   PROCESS Team (UKAEA)
+# Date    :   last modified 2023-07-24
 
 # Specify the minimum version for CMake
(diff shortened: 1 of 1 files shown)


timothy-nunn (approved):
I have verified these changes work on Mac M1 and Ubuntu 20 using both a Python 3.10 and Python 3.8 environment.

PR #2835, opened 2023-07-24
Separate development dependencies

### LINKED ISSUE #2834: Separate development dependencies from required dependencies
Some dependencies are installed (and sometimes version limited) that are only required for process development.
When I install process as a dependency this changes some of the package versions in my environment which is slightly jarring.

Suggested change, separate the developement dependencies into a separate `requirements_dev.txt` file

je-cook:
Closes #2834 

Separates development dependencies into a new file and adds a cmake flag `RELEASE` that will not install those dependencies.

The default install procedure has not changed as most people I'd wager are developing process. 

To use the flag `cmake -S . -B build -DRELEASE=TRUE`


jonmaddock:
Sorry for my delay on this. @timothy-nunn can you take this?

PR #2836, opened 2023-07-27
Instructions for installing PROCESS on CSD3

ym1906:


documentation/proc-pages/usage/csd3-for-process.md
```diff
@@ -0,0 +1,92 @@
...
+- Once you have recieved approval confirmation via email your account will be active.
+- Now you can ssh into CSD3. Note that you must register for two-factor authentication upon your first login request.
+
+    ```bash
+    ssh username@login-cpu.hpc.cam.ac.uk
+    ```
+
+## Install PROCESS on CSD3
+
+PROCESS can be involved using miniconda, a lightweight version of Anaconda which is a distribution of Python and R for scientific computing that simplifies package management.
```
ym1906, line 20: "involved" doesn't make sense, must be a typo - change.

documentation/proc-pages/usage/csd3-for-process.md
```diff
@@ -0,0 +1,92 @@
+# Using the CSD3 service to run PROCESS
+
+[The Cambridge Service for Data Driven Discovery (CSD3)](https://www.hpc.cam.ac.uk/high-performance-computing) is one of the EPSRC Tier2 National HPC Services hosted by Research Computing Services at the University of Cambridge.
+
+This facilty can be used to run batch jobs with PROCESS. This section shall explain how to register for use of the cluster, install PROCESS and submit batch jobs. Find the full documentation for CSD3 [here](https://docs.hpc.cam.ac.uk/hpc/index.html).
+
+## Register to use CSD3
+
+- Register for a [SAFE Account](https://dirac-safe.readthedocs.io/en/latest/safe-guide-users.html#how-to-request-a-dirac-system-account)
```
ym1906, line 9: This link seems to have been removed 26/02/24, update it.

documentation/proc-pages/usage/csd3-for-process.md
ym1906: reolved

documentation/proc-pages/usage/csd3-for-process.md
ym1906: resolved

documentation/proc-pages/usage/csd3-for-process.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+Build then test:
+
+```bash
+cmake -S . -B build
+
+cmake --build build
+
+pytest -k baseline_jan
```
timothy-nunn, line 69:
```suggestion
pytest tests/regression/
```

This appears to be a remnant from closed source process

timothy-nunn (commented):
I'm going to ask for the opinion of @jmorris-uk and @jonmaddock on this one. I agree we should include an example of running on shared compute resources such as CSD3, I don't know if we should be this specific in our **external** documentation. This would be a good internal document, but we might want to consider if it should be in our docs.

If it is, I suspect we shouldn't mention some of the account names.

timothy-nunn (approved):

PR #2840, opened 2023-08-07
CI test job passing despite failures

### LINKED ISSUE #2838: CI test job passing despite failure
The pytest job on the CI is passing when a regression test fails. Clearly this should fail.

jonmaddock:
Closes #2838.


jonmaddock:
In:
```
run: pytest tests/regression -v --reg-tolerance=0 | tee tests/regression/pytest.log || echo "::warning ::Some of the PyTest jobs have failed"
```
The `| tee` and `|| echo` swallow any non-zero exit code returned by `pytest`, resulting in a passing test job when `pytest` itself returns a failure. If we want to `tee` logs in future, we could use `PIPESTATUS` to regain `pytest`'s exit status:
```
$ false | true; echo "${PIPESTATUS[@]}"
1 0
```

I'll remove this for now to get the pipeline failing correctly (i.e. agreement between job and `pytest` status) again.

jonmaddock:
`pytest` is now failing for the unit and regression tests, and this is causing the jobs to fail accordingly, as it should.

PR #2841, opened 2023-08-07
Impurity data files missing

jonmaddock:
The impurity data files have been added, to fix the `FileNotFoundError`s when running the test suite (or Process in general).


jonmaddock:
After adding the impurity data files (which fixed the unit and 5% regression tests `FileNotFoundError`s), I noticed the 0% tolerance regression test was still failing due to differences. After overwriting the "large tokamak" regression test locally, it still failed the test locally. As the references had been overwritten, this might imply non-determinacy: the overwriting run and the subsequent test get different results, or there is a problem with overwriting/testing.

This will be dealt with in #2843. I'll include the impurity data now to get the code functioning again for everyone.

PR #2868, opened 2023-08-30
Add back Stellarator

timothy-nunn:
@stuartmuldrew could you please review the `blanket_neutronics` (https://github.com/timothy-nunn/PROCESS/blob/1855-stellarator-removed-reintroduction-plan/process/stellarator.py#L1034C8-L1034C8) section, as this differs from the original implementation. Other than that, the rest of Stellarator has been put in as-was.


jonmaddock (approved):
Thanks Tim!

PR #2895, opened 2023-09-06
Add issue and PR templates

timothy-nunn:
I think having issue templates will be a good QOL change for this repository pre-open-source. These issue templates are mostly taken from [bluemira](https://github.com/Fusion-Power-Plant-Framework/bluemira).

We should also add a PR template that reflects the proposed format of #2892.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -6,17 +6,8 @@
 
 I confirm that I have completed the following checks:
 
-**Testing**
-- [ ] If I have changed any of the Python or Fortran code, I have rebuilt and run the tests locally and they have passed.
-- [ ] I have addressed any differences in the regression tests caused by this pull request in the comments.
+- [ ] I have justified any large differences in the regression tests caused by this pull request in the comments.
 - [ ] I have added new tests where appropriate for the changes I have made.
 - [ ] If I have had to change any existing unit or integration tests, I have justified this change in the pull request comments.
-
-**Quality**
-- [ ] I have ensured my changes are formatted correctly and conform to PROCESS' quality standards.
 - [ ] If I have made documentation changes, I have checked they render correctly.
-
-**Miscellaneous**
-- [ ] I have created a changelog entry for my change, if appropriate.
 - [ ] I have added documentation for my change, if appropriate.
-- [ ] **I confirm my contribution is made to PROCESS under an MIT license**.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,22 @@
+## Description
+
+<!-- What does this PR do? Please list any issue that these changes address and how you have gone about implementing the changes -->
+
+## Checklist
+
+I confirm that I have completed the following checks:
+
+**Testing**
+- [ ] If I have changed any of the Python or Fortran code, I have rebuilt and run the tests locally and they have passed.
+- [ ] I have addressed any differences in the regression tests caused by this pull request in the comments.
+- [ ] I have added new tests where appropriate for the changes I have made.
+- [ ] If I have had to change any existing unit or integration tests, I have justified this change in the pull request comments.
+
+**Quality**
+- [ ] I have ensured my changes are formatted correctly and conform to PROCESS' quality standards.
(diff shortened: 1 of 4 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,22 @@
+## Description
+
+<!-- What does this PR do? Please list any issue that these changes address and how you have gone about implementing the changes -->
+
+## Checklist
+
+I confirm that I have completed the following checks:
+
+**Testing**
+- [ ] If I have changed any of the Python or Fortran code, I have rebuilt and run the tests locally and they have passed.
```
jonmaddock, line 10: Is this necessary? The CI will do this.

.github/pull_request_template.md
```diff
@@ -0,0 +1,22 @@
...
+- [ ] I have addressed any differences in the regression tests caused by this pull request in the comments.
```
jonmaddock, line 11: I know what you mean, but this is rarely going to be easily explainable. Perhaps this could be changed to "justify large regression test changes" or similar? For discussion.

.github/pull_request_template.md
```diff
@@ -0,0 +1,22 @@
...
+- [ ] I have added new tests where appropriate for the changes I have made.
+- [ ] If I have had to change any existing unit or integration tests, I have justified this change in the pull request comments.
+
+**Quality**
+- [ ] I have ensured my changes are formatted correctly and conform to PROCESS' quality standards.
```
jonmaddock, line 16: This will be addressed by the CI, so this is redundant.

.github/pull_request_template.md
```diff
@@ -0,0 +1,22 @@
...
+- [ ] If I have made documentation changes, I have checked they render correctly.
+
+**Miscellaneous**
+- [ ] I have created a changelog entry for my change, if appropriate.
+- [ ] I have added documentation for my change, if appropriate.
+- [ ] **I confirm my contribution is made to PROCESS under an MIT license**.
```
jonmaddock, line 22: Is this really necessary for every PR? If you're contributing to an MIT-licenced code, isn't this already clearly implied?

.github/pull_request_template.md
```diff
@@ -0,0 +1,22 @@
...
+- [ ] I have addressed any differences in the regression tests caused by this pull request in the comments.
+- [ ] I have added new tests where appropriate for the changes I have made.
+- [ ] If I have had to change any existing unit or integration tests, I have justified this change in the pull request comments.
+
+**Quality**
+- [ ] I have ensured my changes are formatted correctly and conform to PROCESS' quality standards.
+- [ ] If I have made documentation changes, I have checked they render correctly.
+
+**Miscellaneous**
+- [ ] I have created a changelog entry for my change, if appropriate.
```
jonmaddock, line 20: Seeing as hardly anyone has used the changelog tool, is this still worth doing? Perhaps one to think about.

jonmaddock (commented):
These look really good Tim. I've got a few questions on the pull request template, which I feel is a bit much, but maybe this is actually useful/standard and we should be doing it. Let me know your comments.

.github/pull_request_template.md
timothy-nunn: I saw this more as a reminder to check your code *before* pushing it and having it rejected by the CI. I will make this a comment rather than a tickbox I think.

.github/pull_request_template.md
timothy-nunn: This is true, #2892 will make this clear in the contributing guide, so we probably don't need this

.github/pull_request_template.md
timothy-nunn: I know a lot of codes don't really have a CHANGELOG.md anymore and instead use releases to document their changes (which is automatically picked up by GitHub). If we are going to have a changelog we need to enforce its use now we are OS.

But I would be for removing the changelog and keeping the information in the releases

jonmaddock: I agree. I've created #2930 to deal with this issue explicitly.

jonmaddock (approved):
Looks good to me.

PR #2921, opened 2023-09-08
Fetch entire Git history when building docs

timothy-nunn:


timothy-nunn:
Tested on my fork, working

PR #2941, opened 2023-09-12
Get examples.ipynb working

clmould:
## Description

Fixed problems within `examples` folder. The original input file that `examples.ipynb` was trying to use did not exist, so gave it a new input file. The `scan.ipynb` file did not have an input file so one was added. The `csv_output.ipynb` file also did not have an input file, so added one and also changed the list of variables that would be added from the input file to the created csv file. Also added a test for `csv_output.py` which tests that a csv file is created and that the values read in from the input file are all positive floats.

## Checklist


tests/integration/test_examples.py
```diff
@@ -86,3 +86,31 @@ def test_scan(scan_cleanup):
...
+def test_csv(csv_cleanup):
+    """Run the csv_output.py script and check no exceptions are raised.
+
+    csv_output.py intentionally produces files when running the notebook, but remove
+    them when testing.
+    :param csv_cleanup: fixture to delete any produced files
+    :type csv_cleanup: None
+    """
+    # Run entire scan.py script
+    runpy.run_path("csv_output.py")
```
jonmaddock, line 116: Can this perform an assertion on the .csv file contents, like if there are numbers in in?

jonmaddock (changes requested):
Just an assertion addition to the CSV test please!

tests/integration/test_examples.py
```diff
@@ -114,3 +115,16 @@ def test_csv(csv_cleanup):
...
+
+    # Read in the csv file created by test and check it contains values
+    readcsv = pandas.read_csv("csv_output_large_tokamak_MFILE.csv")
+    values = readcsv["Value"]
+    check = False
+    for item in values:
+        check = isinstance(item, float)
+        if check is True:
+            break
+    assert check
```
jonmaddock, line 130: This is much better. But this will pass if any one of the values is a float. Can you check that all the values in the column are floats, without using a for loop?

jonmaddock (changes requested):

tests/integration/test_examples.py
jonmaddock: Hi Clair. What I was trying to get at (and should have explained better) was if this check could be vectorised (i.e. the entire array operated on, rather than looped over one element at a time).

jonmaddock (changes requested):

jonmaddock (approved):
Looks great Clair. A couple of the asserts in the CSV test could be made a bit more concise, but I'm more than happy with everything else in this PR, so I'll merge it. Great!

PR #2956, opened 2023-09-28
Flush convergence callback

je-cook:
## Description

The implemented convergence callback for pyvmcon doesnt flush the print statement therefore in some situations the printing won't appear on stdout until a while later.
This enforces flushing after every update. 

I have no idea how you would test this so I haven't.

## Checklist

### DIFF
--- a/process/solver.py
+++ b/process/solver.py
@@ -169,7 +169,11 @@ class Vmcon(_Solver):
             B = np.identity(numerics.nvar) * self.b
 
         def _solver_callback(i: int, _x, _result, convergence_param: float):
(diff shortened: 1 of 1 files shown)


timothy-nunn (approved):
Cheers James, do we have any idea why this was happening? Was it when PROCESS was being run as a subprocess?

je-cook:
> Cheers James, do we have any idea why this was happening? Was it when PROCESS was being run as a subprocess?

I was running in a subprocess yeah. It sometimes happens if stdout is redirected in some way. Seen it happen on our side too previously when we weren't running a subprocess (in another context not with PROCESS).

PR #2984, opened 2023-11-02
Rearchitect and jit CS fatigue

je-cook:
## Description

I have jitted and reduced recalculation of the same quantities in cs fatigue and reduced calls to np.zeros in pfcoils.
I've also reduced unneeded variable assignment (assigned and then used once) because it is a slow operation in python.
Also * is faster than /

I thought it was worth it for a routine where 50% of time was spent in my run

Speeds up my convergence by >x10 🤯 (2207 iterations -> 134 iterations)

## Checklist


je-cook:
from my side is there any reason to write 1.0e0 as opposed to 1.0. I would suggest the latter is clearer if there isn't a reason

ajpearcey (approved):
Looks good to me

timothy-nunn (approved):

PR #2985, opened 2023-11-02
Use for loops instead of while loops if possible

je-cook:
## Description

While loops can be slow so I've replaced most of them:

https://stackoverflow.com/questions/68314464/python-performance-while-vs-for-loops

## Checklist

### DIFF
--- a/process/pfcoil.py
+++ b/process/pfcoil.py
@@ -2973,10 +2973,8 @@ class PFCoil:
 
             # Actual current density in superconductor, which should be equal to jcrit(thelium+tmarg)
             # when we have found the desired value of tmarg
-            lap = 0
-            while True:
-                lap = lap + 1
-                if (ttest <= 0.0) or (lap > 100):
+            for lap in range(100):
+                if ttest <= 0.0:
                     eh.idiags[0] = lap
                     eh.fdiags[0] = ttest
                     eh.report_error(158)
@@ -3006,6 +3004,10 @@ class PFCoil:
                     jcritp = jcritp + (jcritp * 1e-6)
 
                 ttest = ttest - 2.0e0 * delt * (jcrit0 - jsc) / (jcritp - jcritm)
+            else:
+                eh.idiags[0] = lap
+                eh.fdiags[0] = ttest
+                eh.report_error(158)
 
             tmarg = ttest - thelium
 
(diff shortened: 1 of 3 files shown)


process/vacuum.py
```diff
@@ -401,12 +401,10 @@ def vacuum(
             ceff[i] = 1.0e0 / (nduct / s[i] - 1.0e0 / (sp[i] * pumpn))
 
             #  Newton's method solution for duct diameter
-
             while True:
                 d[i] = 1.0e0
 
-                lap = 0
-                while True:
+                for _ in range(99):
```
timothy-nunn, line 407: I think this should be 100.

process/pfcoil.py
```diff
@@ -2973,10 +2973,8 @@ def deltaj_hijc_rebco(temperature):
 
             # Actual current density in superconductor, which should be equal to jcrit(thelium+tmarg)
             # when we have found the desired value of tmarg
-            lap = 0
-            while True:
-                lap = lap + 1
-                if (ttest <= 0.0) or (lap > 100):
```
timothy-nunn, line 2979: Again because of the > my checks are showing this loop might run one less than before (in practice it doesn't matter), but I think retaining the same behaviour would be ideal. Let me know if im being silly

timothy-nunn (commented):
I think some of the loops might both run one less time in extreme cases

process/pfcoil.py
je-cook: it will run from 0 -> 99 (100 runs) and previously because immediately `lap = 1`  this will run from 1 to 100 (100 runs) and then on the 101st immediately break.

I think thats right...

process/vacuum.py
je-cook: this one you may be right

process/pfcoil.py
je-cook: eg
```python
l = 0

while True:
    l += 1
    if l > 5:
        break
    print("hi")
```
will print 5 'hi's

process/vacuum.py
je-cook: yeah this one should be 100 because of the placement of the break

process/pfcoil.py
timothy-nunn: Yep, logic is after the `break` on this one, you're right

timothy-nunn (approved):

Issue #2998, opened 2023-11-24
Update heating and current drive documentation to demonstrate all models and variables.


chris-ashe:
The documentation for the heating and current drive systems is currently only one page and does not describe the operation of all the models and how to properly set up the variables

PR #2999, opened 2023-11-24
2994 new eccd current drive model to account for zeff and te with a cutoff function

chris-ashe:
## Description

New model has been input to account for ECCD cutoff as a function of electron temperature and effective charge

- [x] Dependance of the harmonic number has been verified
- [x] Dependance between O and X mode has been verified
- [x] Quantified the difference between `effrfss` & `effcd`
- [x] ECRH HARE model seems to be removed. Should tidy up
- [x] Update plotting dependencies eg `plot_proc`.

## Checklist

### DIFF
--- a/process/current_drive.py
+++ b/process/current_drive.py
@@ -710,6 +710,11 @@ class CurrentDrive:
             )
         elif current_drive_variables.iefrf == 12:
             po.ocmmnt(self.outfile, "EBW current drive")
+        elif current_drive_variables.iefrf == 13:
+            po.ocmmnt(
+                self.outfile,
+                "Electron Cyclotron Current Drive (O-mode cutoff with Zeff & Te)",
+            )
 
         po.ovarin(
             self.outfile,
@@ -732,6 +737,11 @@ class CurrentDrive:
             )
         elif current_drive_variables.iefrffix == 12:
             po.ocmmnt(self.outfile, "EBW current drive")
+        elif current_drive_variables.iefrffix == 13:
+            po.ocmmnt(
+                self.outfile,
+                "Electron Cyclotron Current Drive (O-mode cutoff with Zeff & Te)",
+            )
 
         po.ovarin(
             self.outfile,
--- a/process/current_drive.py
+++ b/process/current_drive.py
@@ -194,8 +194,54 @@ class CurrentDrive:
                 )
 
                 effcdfix = effcdfix * density_factor
-
                 effrfssfix = effrfssfix * density_factor
+            elif current_drive_variables.iefrffix == 13:
+                # ECCD model for O-mode cut-off with added Te and Zeff dependance
+                # Scaling author: Simon Freethy
+                # Ref : PROCESS issue #2994
+
+                fc = (
+                    1
+                    / (2 * np.pi)
+                    * constants.echarge
+                    * physics_variables.bt
+                    / constants.emass
+                )
+                fp = (
+                    1
+                    / (2 * np.pi)
+                    * np.sqrt(
+                        (
+                            (physics_variables.dene / 1.0e19)
+                            * constants.echarge**2
+                            / (constants.emass * constants.epsilon0)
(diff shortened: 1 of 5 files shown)


process/current_drive.py
```diff
@@ -226,7 +286,7 @@ def cudriv(self, output: bool):
                     * 1.0e6
                 )
                 faccdfix = auxiliary_cdfix / physics_variables.plascur
-            elif current_drive_variables.iefrffix in [3, 7, 10, 11, 12]:
```
timothy-nunn, line 229: Why was `11` removed here? Was it a mistake?

process/current_drive.py
```diff
@@ -403,11 +463,100 @@ def cudriv(self, output: bool):
...
+                )
+
+                density_factor = 0.5e0 * (
+                    1.0e0 + np.tanh((2.0e0 / a) * ((fp - fc) / fp - a))
+                )
+
+                current_drive_variables.effcd = (
+                    current_drive_variables.effcd * density_factor
+                )
+                effrfss = effrfss * density_factor
```
timothy-nunn, line 497: 470 to this line doesn't make sense to me. Why have we changed how the old switch behaves in this PR?

process/current_drive.py
```diff
@@ -721,13 +880,27 @@ def cudriv(self, output: bool):
                 "(gamma_ecrh)",
                 current_drive_variables.gamma_ecrh,
             )
-        elif current_drive_variables.iefrf == 12:
+        if current_drive_variables.iefrf == 12:
```
timothy-nunn, line 883: why?

process/current_drive.py
```diff
@@ -721,13 +880,27 @@ def cudriv(self, output: bool):
...
                 current_drive_variables.xi_ebw,
             )
+        if current_drive_variables.iefrf in [12, 13]:
+            po.ovarre(
+                self.outfile,
+                "EC harmonic number",
+                "(harnum)",
+                current_drive_variables.harnum,
+            )
+        if current_drive_variables.iefrf == 13:
```
timothy-nunn, line 897: Would more 'correct' to move this block under the `elif current_drive_variables.iefrf == 12:` and have it as `elif current_drive_variables.iefrf == 13:` and then have the if in block below as is.

timothy-nunn (changes requested):
Couple of questions that need addressing.

process/current_drive.py
chris-ashe: The HARE model that we used to have had a switch value of 11. The model was removed but its switch remnants still remained in some places. Was tidying up

process/current_drive.py
chris-ashe: Realised that the cutoff function section was commented out so added it back in. The cutoff function was in for the `iefrffix` but not `iefrf`

process/current_drive.py
timothy-nunn: Ah ok. Could you note in the docs that 11 has been removed?

timothy-nunn:
Also @chris-ashe, if you have some reference values would it be possible to add in a quick test for this

process/current_drive.py
chris-ashe: I am updating all of the current drive docs in https://github.com/ukaea/PROCESS/issues/2998. Which captures this change

process/current_drive.py
chris-ashe: It doesn't really matter in the order of the output. At the moment the output goes efficiency scaling value -> harmonic number -> wave mode. `iefrf = 13` dose not a efficiency scaling value.

chris-ashe:
Test is now covered in #3013

jonmaddock (approved):
Looks great after rebasing.

jonmaddock:
As Tim's questions have been answered and Chris is keen to get this in, I hope you don't mind me merging this Tim!

PR #3005, opened 2023-12-04
3004 remove remnants of plasmod functionality

chris-ashe:
## Description

Some bits of PLASMOD and the associated `ipedestal = 2,3` options were still present these have been removed.
@timothy-nunn PLASMOD is still noted in example.ipynb
## Checklist


ajpearcey (approved):
looks good

process/io/plot_proc.py
```diff
@@ -2807,8 +2716,7 @@ def parse_args(args):
     # Setup command line arguments
     parser = argparse.ArgumentParser(
         description="Produces a two page summary of the PROCESS MFILE output, using the MFILE.  "
-        "For info contact michael.kovari@ukaea.uk or james.morris2@ukaea.uk.  "
-        "If using PLASMOD you must specify the profile file using -m."
+        "For info contact michael.kovari@ukaea.uk or james.morris2@ukaea.uk."
```
timothy-nunn, line 2719: This would be a good opportunity to point people to issues/discussion/the list of contacts rather than using email (that may one day be outdated).

process/utilities/errorlist.json
```diff
@@ -878,12 +878,12 @@
     {
       "no": 174,
       "level": 3,
-      "message": "CONVERT_PLASMOD2PROCESS: PLASMOD Transport model (ipedestal=3) has not converged"
+      "message": "OBSOLETE"
```
timothy-nunn, line 881: I suspect we can actually removed these entries from the json file entirely since they should never be called, and if they were it would raise an error (probably what we want).

timothy-nunn (changes requested):
One thing I think should be changed, another thing that can be done while your doing that.

process/io/plot_proc.py
chris-ashe: Changed to the contact list on the front page

timothy-nunn (approved):

PR #3008, opened 2023-12-05
fix bug in output filename calculation in plot_radial_build.py

alekitto:
## Description

Filename calculation in `plot_radial_build.py` is done taking the last 10 characters of the input file path.
This behaviour is not safe and can lead to `NotFoundError` to be thrown or cause path traversal in the worst case.

The fix replaces the unsafe part picking up the basename of the file (without extension) and prepending it to `_radial_build.{save_format}` static part.

This also fixes `test_plot_radial_build` run is some environments.

## Checklist

### DIFF
--- a/process/io/plot_radial_build.py
+++ b/process/io/plot_radial_build.py
@@ -460,7 +460,7 @@ def main(args=None):
     plt.xlabel("Radius [m]")
     plt.tight_layout()
     plt.savefig(
-        f"{args.outputdir}/{args.input[:-10]}_radial_build.{save_format}",
+        f"{args.outputdir}/{Path(args.input).stem}_radial_build.{save_format}",
         bbox_inches="tight",
     )
 


process/io/plot_radial_build.py
```diff
@@ -460,7 +461,7 @@ def main(args=None):
     plt.xlabel("Radius [m]")
     plt.tight_layout()
     plt.savefig(
-        f"{args.outputdir}/{args.input[:-10]}_radial_build.{save_format}",
+        f"{args.outputdir}/{os.path.splitext(os.path.basename(args.input))[0]}_radial_build.{save_format}",
```
timothy-nunn, line 464:
```suggestion
        f"{args.outputdir}/{Path(args.input).stem}_radial_build.{save_format}",
```

This is probably the correct and less-verbose way to do this in newer versions of Python (of course this would also make the `import os` redundant.

timothy-nunn (changes requested):
Hi @alekitto,

Thank you for your contribution to the PROCESS repository by identifying and proposing a fix to this bug. 

I think pathlib may offer a nicer, more modern fix to this bug, and have suggested as such.

process/io/plot_radial_build.py
alekitto: Yes, you're right. Fixed

timothy-nunn: Thank you very much :)

timothy-nunn (approved):

PR #3024, opened 2024-01-18
fix wrongly reporting multiple obsolete vars when only one is present in input file

### LINKED ISSUE #3020: [BUG] Multiple obsolete variables reported when only one is present in input file
## Describe the bug

I recently tried to run an input file `test_IN.DAT` (called test_IN.log as github doesn't let me upload .DATs) 
[test_IN.log](https://github.com/ukaea/PROCESS/files/13901141/test_IN.log).
Process reported: 
[test.log](https://github.com/ukaea/PROCESS/files/13901150/test.log)
But only the first error variable was actually obsolete (`sigvvall` should be changed to `max_vv_stress`). Process seemed to report every variable in the input file after the actual obsolete variable as obsolete.

## Expected behaviour

Process should only report the actual obsolete variables, and hint the new variable name (this should be added to the `obsolete_vars.py` dict too).

## Environment

- Version: v3.0.0

clmould:
## Description

Closes #3020 
- added missing obsolete variables to obsolete_vars.py
- added ValueError to report obsolete variables to replace STOP 1

## Checklist


process/main.py
```diff
@@ -521,21 +521,44 @@ def validate_input(self):
...
-        if len(replace_hints) > 0:
-            print(
-                "The IN.DAT file contains obsolete variables from the OBS_VARS dictionary."
-            )
-            print(
-                f"The obsolete variables in your IN.DAT file are: {obs_vars}. Please change them to the following corresponding new variable(s) before continuing: {replace_hints}."
+            if var in obsolete_variables:
+                obs_vars_in_in_dat.append(var)
+                new_var = obsolete_variables.get(var)
+                replace_hints[var] = new_var
```
timothy-nunn, line 530: These two lines are a bit confusing and I think thats probably because of the the variable names. I think you could just do the following:

```suggestion
                replace_hints[var] = obsolete_variables.get(var)
```

process/main.py
```diff
@@ -521,21 +521,44 @@ def validate_input(self):
...
+                "Either remove these or replace them with their updated variable names. "
             )
+            for var in obs_vars_in_in_dat:
+                if replace_hints[var][0] is None:
+                    message = (
+                        message
+                        + "\n \n "
+                        + var
+                        + " is an obsolete variable and does not have a replacement. "
+                    )
```
timothy-nunn, line 545: Could probably be a one-line f-string

```suggestion
                    message += f"\n\n {var} is an obsolete variable and does not have a replacement."
```

Also the message feels a bit clunky. Maybe you could say "... is an obsolete variable and should be removed"

process/main.py
```diff
@@ -521,21 +521,44 @@ def validate_input(self):
...
+                if replace_hints[var][0] is not None:
+                    message = (
+                        message
+                        + "\n \n "
+                        + var
+                        + " is an obsolete variable and needs to be replaced by "
+                        + str(replace_hints[var][0])
+                        + ". "
+                    )
+                    if len(replace_hints[var]) == 2:
```
timothy-nunn, line 557: What about if I don't provide a replacement but provide a help message?

process/main.py
```diff
@@ -521,21 +521,44 @@ def validate_input(self):
...
+                if replace_hints[var][0] is None:
+                    message = (
+                        message
+                        + "\n \n "
+                        + var
+                        + " is an obsolete variable and does not have a replacement. "
+                    )
+                    if len(replace_hints[var]) == 2:
+                        message = message + str(replace_hints[var][1])
+                if replace_hints[var][0] is not None:
```
timothy-nunn, line 548:
```suggestion
                else:
```

source/fortran/input.f90
```diff
@@ -1213,18 +1145,6 @@ subroutine parse_input_file(in_file,out_file,show_changes)
...
 
-       case ('target_spread', 'lambda_q_omp', 'lcon_factor', 'netau_sol', 'kallenbach_switch', &
-         'kallenbach_tests', 'kallenbach_test_option', 'kallenbach_scan_switch', 'kallenbach_scan_var', &
-         'kallenbach_scan_start', 'kallenbach_scan_end', 'kallenbach_scan_num', 'targetangle', 'ttarget', &
-         'qtargettotal', 'impurity_enrichment', 'fractionwidesol', 'abserr_sol', 'relerr_sol', 'mach0', &
-         'neratio')
-         write(outfile,*) ' '
-         write(outfile,*) '**********'
-         write(outfile,*) 'The Stellarator model is currently not included in PROCESS.'
-         write(outfile,*) 'See issue #1853 for more information on the use of Stellarators.'
```
timothy-nunn, line 1224: I see where the above confusion came from! I made a mistake here :/

timothy-nunn (changes requested):
I like the idea of consolidating the help message and hints into `obsolete_vars.py` however I think the logic that generates the error messages could be improved to provide more flexibility when defining a obsolete variable that has no replacement but has a help message.

In fact, I'm not so sure `stellarator_message` would be printed out if you added one of the obsolete variables (from the Kallenbach model) to it.

timothy-nunn (approved):

PR #3037, opened 2024-01-30
1541 calling of avail from stellaratorife does not take into account iavail

chris-ashe:
## Description

As per discussion in !601, IFE/Stellarator call the availability model. However, they both call the avail subroutine without taking into account the iavail flag. However, other calls to availability models take into account the iavail flag:
```
if cv.iavail > 1:
    self.avail_2(output)  # Morris model (2015)
else:
    self.avail(output)  # Taylor and Ward model (1999)
```
Upon further inspection it has been seen that `avail` is specifcally for IFE use so has not been changed.
Though for the stellarator model any of the avaialbility models can be used as the same components are used between the model (tokamak) and the stellarator.
The stellarator module now runs `availability.run(output=True)` to allow the user to specify the `avail` switch for their preferred model.

## Checklist

### DIFF
--- a/process/ife.py
+++ b/process/ife.py
@@ -48,8 +48,6 @@ class IFE:
             self.costs.costs(output=True)
 
             # Plant availability
-            # TODO: should availability.run be called
-            # rather than availability.avail?
             self.availability.avail(output=True)
 
             # IFE physics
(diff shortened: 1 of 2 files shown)


j-a-foster (approved):
The TODO has been correctly done and matches calling the same model elsewhere in the code.

PR #3038, opened 2024-01-30
Add actual constraint numbers to mfile

jonmaddock:
Enables identification of constraints: rather than having `normres001, normres002, ..., constr007, constr008`, use the actual constraint numbers, e.g. `eq_constr001, eq_constr007, ..., ineq_constr010, ineq_constr009`. This allows the constraint to be identified from the `MFILE.DAT` alone, instead of needing to cross-check with the `IN.DAT` to find that `constr007` is actually constraint number 10. This also preserves the order in which the constraints are defined in the `IN.DAT`.


ajpearcey (approved):
Happy for this to be merged.

PR #3040, opened 2024-01-30
Constraint type bug in in dat.py

jonmaddock:
When creating a new `IN.DAT` file from an existing `MFILE.DAT`, the constraints are being ordered numerically. This is unwanted if fixing f-values so that equality and inequality constraints are separated, rather than all being equalities, as they become mixed up. 

Preventing this re-ordering preserves what's eq and what's ineq when using the `neqns` variable; the first `neqns` are eqs, rest are ineqs. This is now preserved from `MFILE.DAT` to `IN.DAT`.


ajpearcey (approved):
Happy for this to be merged.

PR #3044, opened 2024-02-02
Add objective function to once-through mfile

jonmaddock:
The normalised objective function is missing from once-through mfile output.


ajpearcey (approved):
Happy for this to be merged.

PR #3053, opened 2024-02-14
Added check to see if fimp(2) deviates from default value, if it does…

ym1906:
… write a warning message to use ralpne instead.

## Description

This value is controlled by ralpne , in fact changing fimp(2) does not change the value of the parameter. It also doesn't make sense that a user would set these two parameters to different values in the first place.

However, as fimp(2) is part of the fimp array, it's not straight forward to remove it as an input variable. All members of fimp(1:13) are inputs, and all of them are output (and still used by other parts of the code). A generic fortran function loop for looping through arrays is used to set them as inputs. It doesn't seem elegant to add a specific exception to this function.

I have added a warning and stop to the user when they change fimp(2) from the default value, and they are told to use ralpne instead.

## Checklist

### DIFF
--- a/source/fortran/initial.f90
+++ b/source/fortran/initial.f90
@@ -347,6 +347,11 @@ subroutine check
         trithtmw = 0.0D0
     end if
 
+    if (fimp(2) .ne. 0.1D0) then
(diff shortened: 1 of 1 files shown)


ajpearcey (approved):
lgtm

PR #3054, opened 2024-02-14
Correct nviter and convergence_parameter reporting in output

timothy-nunn:
## Description

PyVMCON interface in PROCESS now updates the `nviter` and `convergence_parameter` values so they can be correctly reported in the OUT.DAT and MFILE.DAT

I have also corrected the dummy variables in the `_solver_callback` because they were the wrong way around.

And, I have removed `nviter` from the regression test comparison because it does not make much sense to fail if the number of iterations has changed (we should only care about the solution).

### DIFF
--- a/process/solver.py
+++ b/process/solver.py
@@ -169,7 +169,9 @@ class Vmcon(_Solver):
         if self.b is not None:
             B = np.identity(numerics.nvar) * self.b
 
-        def _solver_callback(i: int, _x, _result, convergence_param: float):
+        def _solver_callback(i: int, _result, _x, convergence_param: float):
+            numerics.nviter = i + 1
+            global_variables.convergence_parameter = convergence_param
             print(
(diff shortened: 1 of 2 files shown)


timothy-nunn:
Large tokamak now reports the below in the OUT.DAT:

```
VMCON error flag                                                         (ifail)                           1     
 Number of iteration variables                                            (nvar)                           45     
 Number of constraints (total)                                            (neqns+nineqns)                  26     
 Optimisation switch                                                      (ioptimz)                         1     
 Figure of merit switch                                                   (minmax)                          1     
 Square root of the sum of squares of the constraint residuals            (sqsumsq)                 1.871E-04  OP 
 VMCON convergence parameter                                              (convergence_parameter)   4.087E-09  OP 
 Normalised objective function                                            (norm_objf)               1.600E+00  OP 
 Number of VMCON iterations                                               (nviter)                          8  OP 
```

The MFILE also agrees

jonmaddock (approved):

PR #3057, opened 2024-02-14
Expand bounds of fhfac to reduce chance power balance cannot be found

timothy-nunn:
## Description

Changed the bounds when trying to find the zero of `fhz` to reduce the chance there is no 0 point within the bounds (current `[0.01, 100]`


chris-ashe (approved):
LGTM

PR #3060, opened 2024-02-15
736 warnings for function ncore

chris-ashe:
## Description
The logging of the ncore error to the terminal during iterations has been removed and replaced with a level 2 error that is shown at the end of the run. This warns the user to raise the lower bound of dene to prevent VMCON going near negative calculated values of n0. This is only shown when the run is finished. 

Documentation has been added warning the user to use constraint 81 to prevent un-realistic profiles and to be weary about setting the lower bound of dene too low. 

The bounds in `iteration_variables.f90`  for `fgwped, fgwsep & dene` have been varied to represent the values found in realistic solutions.

Long term solution is being discussed in #3059 

## Checklist

### DIFF
--- a/process/profiles.py
+++ b/process/profiles.py
@@ -157,6 +157,7 @@ class NProfile(Profile):
             # Allows solver to continue and
             # warns the user to raise the lower bound on dene if the run did not converge
             error_handling.report_error(282)
+            ncore = 1.0e-6
         return ncore
 
     def set_physics_variables(self):
--- a/process/profiles.py
+++ b/process/profiles.py
@@ -154,9 +154,8 @@ class NProfile(Profile):
         )
 
         if ncore < 0.0:
-            # Prevent ncore from going negative (and terminating the optimisation) by
-            # kludging to small positive value. Allows solver to continue and
-            # hopefully be constrained away from this point (e.g. constraint 81, ne0 > neped)
+            # Allows solver to continue and
+            # warns the user to raise the lower bound on dene if the run did not converge
             error_handling.report_error(282)
         return ncore
 
--- a/process/profiles.py
+++ b/process/profiles.py
@@ -3,10 +3,7 @@ import logging
 from scipy import integrate
 from abc import ABC, abstractmethod
 
-from process.fortran import (
-    maths_library,
-    physics_variables,
-)
+from process.fortran import maths_library, physics_variables, error_handling
 
 logger = logging.getLogger(__name__)
 # Logging handler for console output
@@ -159,12 +156,9 @@ class NProfile(Profile):
             # Prevent ncore from going negative (and terminating the optimisation) by
             # kludging to small positive value. Allows solver to continue and
             # hopefully be constrained away from this point (e.g. constraint 81, ne0 > neped)
-            logger.exception("ncore is negative. Kludging to 1e-6.")
-            ncore = 1.0e-6
-            # raise ValueError(
-            #     f"Core density is negative: {ncore}. {nped = }, {nsep = }, {nav = }"
-            # )
-        return ncore
+            print("hello")
+            error_handling.report_error(252)
+        return
 
     def set_physics_variables(self):
         """_summary_
--- a/process/profiles.py
+++ b/process/profiles.py
@@ -124,7 +124,9 @@ class NProfile(Profile):
         )
 
     @staticmethod
-    def ncore(rhopedn, nped, nsep, nav, alphan):
+    def ncore(
+        rhopedn: float, nped: float, nsep: float, nav: float, alphan: float
+    ) -> float:
         """This routine calculates the core denesity of a pedestalised profile.
 
         :param rhopedn: normalised minor radius pedestal position
@@ -156,8 +158,9 @@ class NProfile(Profile):
             # Prevent ncore from going negative (and terminating the optimisation) by
             # kludging to small positive value. Allows solver to continue and
             # hopefully be constrained away from this point (e.g. constraint 81, ne0 > neped)
-            print("hello")
             error_handling.report_error(282)
+
+        return ncore
 
     def set_physics_variables(self):
         """_summary_
--- a/process/profiles.py
+++ b/process/profiles.py
@@ -152,13 +152,12 @@ class NProfile(Profile):
             )
         )
 
-        if ncore < 0:
+        if ncore < 0.0:
             # Prevent ncore from going negative (and terminating the optimisation) by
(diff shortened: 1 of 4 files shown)


mkovari:
I would still like the warning to be explicit about what sort of lower and upper bounds people should set.  

For example:
`dene` Recommended value of `boundl(6)` > 80% of expected volume-averaged density.

It *might* also help to narrow the default bounds as follows.

### Separatrix density
https://github.com/ukaea/PROCESS/blob/b8c0a3d2259b3c8541011960f6ae3c405274da6d/source/fortran/iteration_variables.f90#L3378-L3380  Change to: 
```
    lablxc(152) = 'fgwsep        '
    boundl(152) = 0.001D0
    boundu(152) = 0.1D0
```
### Pedestal density
```
    lablxc(145) = 'fgwped        '
    boundl(145) = 0.500D0
    boundu(145) = 1.000D0
```
Change to 
```
    lablxc(145) = 'fgwped        '
    boundl(145) = 0.100D0
    boundu(145) = 0.75D0
```

```
    lablxc(6) = 'dene          '
    boundl(6) = 1.00D19
    boundu(6) = 1.00D21
```
Change to
```
    lablxc(6) = 'dene          '
    boundl(6) = 2.00D19
    boundu(6) = 1.00D21
```

source/fortran/iteration_variables.f90
```diff
@@ -3377,7 +3377,7 @@ subroutine init_itv_152
     implicit none
     lablxc(152) = 'fgwsep        '
     boundl(152) = 0.001D0
-    boundu(152) = 1.000D0
+    boundu(152) = 0.1D0
   end subroutine init_itv_152
```
timothy-nunn, line 3381: I am a little concerned about changing the default bounds as this could lead to existing input files (that use the default bounds) starting outside of the feasible region. In this instance, the 3 changes are minor so I think I am happy with it.

timothy-nunn (commented):
@chris-ashe if you answer my comment I will approve it subject to your reply

process/profiles.py
```diff
@@ -155,15 +153,10 @@ def ncore(rhopedn, nped, nsep, nav, alphan):
...
-            # hopefully be constrained away from this point (e.g. constraint 81, ne0 > neped)
-            logger.exception("ncore is negative. Kludging to 1e-6.")
-            ncore = 1.0e-6
-            # raise ValueError(
-            #     f"Core density is negative: {ncore}. {nped = }, {nsep = }, {nav = }"
-            # )
+        if ncore < 0.0:
+            # Allows solver to continue and
+            # warns the user to raise the lower bound on dene if the run did not converge
+            error_handling.report_error(282)
```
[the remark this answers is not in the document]
chris-ashe: Running the files it dosent seem to make a difference as they all fail eventually anyway. It seems better to have it fail quickly and the user updates the lower bound to prevent it going negative again in future runs

source/fortran/iteration_variables.f90
chris-ashe: The output values in the large tokamak regression are 0.85 and 0.5 for pedestal and separatrix respectively. @mkovari 
Which is outside the bounds recommended

process/profiles.py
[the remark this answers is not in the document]
timothy-nunn: I probably need @jonmaddock to comment on this, as he is the one that wrote the kludge. For now, since this is a level 2 error, I will approve it but will make a follow-up issue to check that this fix is ok with the UQ tools (which is why the kludge was added)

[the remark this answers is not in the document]
chris-ashe: If needs be we could put the kludge back in without adding back the logger output as it just slows down the terminal output

source/fortran/iteration_variables.f90
chris-ashe: Electron density at separatrix / nGW                                     (fgwsep_out)              5.000E-01
Electron density at pedestal / nGW                                       (fgwped_out)              8.500E-01

From the output file of the large tokamak regression test

mkovari: I would still like the warning to be explicit about what sort of lower and upper bounds people should set.

For example:
dene Recommended value of boundl(6) > 80% of expected volume-averaged density.

process/profiles.py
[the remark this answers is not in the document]
timothy-nunn: Probably a good idea, and then #3059 can address making this model more robust

source/fortran/iteration_variables.f90
chris-ashe: Is this true for all cases? Even for a stellerator?
The output warning are limited to one line so I will need to amend the current one as I cant add it under unless we change their length

timothy-nunn: Maybe point people to a section in the documentation with the warning? And have a longer explanation in the docs.

mkovari:
With all due respect, I find this approach slightly annoying.  Wouldn't it be best to agree first on what should be done?

chris-ashe:
@mkovari  This is a temporary fix for papercuts and a more robust solution can be discussed in #3059

mkovari:
So what has actually changed?

chris-ashe:
- `Kluding ncore` is no longer spammed to the terminal when PROCESS is iterating.
- A level 2 error is raised if `ncore` now goes negative telling the user to raise the value of `dene` and or its lower bound.
- Updates to the pedestal documentation about how to set up a file with constraint 81 and the correct bounds of `dene ` to prevent ncore going negative

chris-ashe:
- `Kluding ncore` is no longer spammed to the terminal when PROCESS is iterating.
- A level 2 error is raised if `ncore` now goes negative telling the user to raise the value of `dene` and or its lower bound.
- Updates to the pedestal documentation about how to set up a file with constraint 81

timothy-nunn:
I agree that the solution to this issue and the papercut is to remove the warning "spam" and providing help in the documentation. Subsequent discussions can decide a long-term solution to our profiles and will be able to consider how best we can stop unphysical profiles from occurring.

timothy-nunn (approved):

chris-ashe:
@mkovari  Are you happy with the changes as of now that provide a route to solving the negative problems by pointing to the documentation and raising new warning?
We can discuss a longer term and more robust solution with @jonmaddock that will work in a range of different scenarios that doesn't affect convergence.

mkovari:
I am happy in the sense that I don't think these changes are going to make the situation any worse.

PR #3066, opened 2024-02-20
609-vdalw-and-fvdump

### LINKED ISSUE #609: vdalw and fvdump
In GitLab by @jmorris-uk on Feb 7, 2018, 13:57

Is there a reason why both `vdalw` and `fvdump` are iteration variables. Having looked at Git blame it appears they were both added 63 months ago by `pknight`.

Normally for a constraint equation only `fvdump` would be an iteration variable and `vdalw` would be a user input? Should we remove `vdalw` as an iteration variable?

@mkovari @rkemp @hlux

ajpearcey:
closes #609


timothy-nunn (approved):
Solution does not appear to change too much following this PR.

PR #3084, opened 2024-02-29
Implement cost manager for cost model calling

timothy-nunn:
## Description

Calls the cost models via a manager that will allow the resolving of the concrete cost model to be deferred until the caller first makes a call to the model/ another model calls the costs. This is relevant to making sure the Stellarator can take a reference to the cost model. 

This code does make the cost model calling and resolution a little bit complex but seems to be the best solution to allow flexible cost models.

## Checklist

### DIFF
--- a/process/main.py
+++ b/process/main.py
@@ -567,7 +567,7 @@ class CostsProtocol(Protocol):
         """write model output"""
 
 
-class CostManager(CostsProtocol):
+class CostManager:
     def __init__(self, models):
         """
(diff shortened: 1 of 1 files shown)


process/main.py
```diff
@@ -568,6 +567,34 @@ def output(self):
         """write model output"""
 
 
+class CostManager(CostsProtocol):
```
je-cook, line 570: I dont think the CostManager should inherit from the protocol
```suggestion
class CostManager:
```

PR #3102, opened 2024-03-14
Replaced local calculated eps variable with fortran variable as input

### LINKED ISSUE #454: eps - inverse aspect ratio
In GitLab by @hlux on Dec 2, 2016, 10:26

eps is defined as the inverse aspect ratio in global variables (typically denominated by Greek letter epsilon in physics literature). However, eps seems to be also used as local variable in several occasions throughout the PROCESS code. 
It should be checked that where ever the local variable should be used, the global variable is not imported. (I assume this is true, but just wanted to be sure.)

This should also not be an issue anymore once we have proper encapsulation and clear interfaces. So maybe it is not urgent to check this.

j-a-foster:
Replaced inverse aspect ratio calculated locally in `_plascar_bpol` with input `eps` which is the fortran variable `physics_variables.eps` in subsequest functions which use `_plascar_bpol`. These functions are `bpol` and `plasc`.

Closes #454 

## Checklist


j-a-foster:
Unsure whether need to change [physics.py line 751](https://github.com/ukaea/PROCESS/pull/3102/files#diff-e21acd338af89777d2e84c823a6854ec4e80ca6fcf7ebb8fe062fbfc36d51c9cR751) as it's calculated on an index by index basis. It's also not clear from the docstring/comments whether this `eps` corresponds to the `eps` mentioned in #454.

mkovari (approved):

PR #3107, opened 2024-03-18
updated q definition

chris-ashe:
## Description

Variable descriptions have been updated and or added to give clearer definition of `q`

## Checklist

### DIFF
--- a/source/fortran/input.f90
+++ b/source/fortran/input.f90
@@ -704,7 +704,7 @@ contains
                'Plasma resistivity pre-factor')
        case ('q')
           call parse_real_variable('q', q, 1.00D0, 50.0D0, &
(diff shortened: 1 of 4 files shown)


timothy-nunn (approved):
I agree with https://github.com/ukaea/PROCESS/discussions/3106 that we need to improve our definition of `q` in PROCESS. The presence of `q` and `q95` makes everything a little complex. 

However, I also agree for the need to carefully consider how we do this in a backwards compatible way, hence why this PR should just improve the visibility of the definition without attempting to rewrite how PROCESS handles these variables.

source/fortran/input.f90
mkovari, line 1: I think "near plasma edge"is a little vague.

PR #3115, opened 2024-03-26
Warning when profile parameters are reset in parabolic_paramterisation

timothy-nunn:
Outputs the following warning if the parabolic parameterisation is inappropriately used

> process.plasma_profiles - WARNING - Parabolic plasma profiles is used for an L-Mode plasma, but the physics variables do not describe an L-Mode plasma. 'rhopedt', 'rhopedn', 'teped', 'tesep', 'neped', 'nesep', and 'tbeta' have all been reset to L-Mode appropriate values

Tested by using a modified large-tokamak input file with `ipedestal=0`, the warning is output.

### DIFF
--- a/process/plasma_profiles.py
+++ b/process/plasma_profiles.py
@@ -58,14 +58,30 @@ class PlasmaProfile:
 
     def parabolic_paramterisation(self):
         """Parameterise plasma profiles in the case where ipedestal=0"""
-        #  Reset pedestal values to agree with original parabolic profiles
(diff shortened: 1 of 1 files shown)


mkovari:
It might be good to mention explicitly that `ipedestal=0` in the error message.

You might also change `parabolic_paramterisation` to 'parabolic_parameterisation' if you are really keen.

ajpearcey (approved):
lgtm

PR #3116, opened 2024-03-26
782 simple model to calculate the allowable current density in a superconducting coil

### LINKED ISSUE #782: "Simple model to calculate the allowable current density in a superconducting coil"
In GitLab by @mkovari on Nov 19, 2018, 14:30

Subroutine `sctfjalw` is described as follows.
>  This routine using a simple model to calculate the allowable current density in a superconducting coil, given the magnetic field and the allowable stress.

I don't think anyone knows what it does and I don't think we have ever used it.  I propose we remove this subroutine.  I will remove it unless I hear otherwise.

This *superconducting* model is only called when `tfc_model = 0`.  However, this switch is defined as follows:
```
tfc_model /1/ : switch for TF coil magnet stress model:
                = 0 simple model (solid copper coil)
                = 1 CCFE two-layer stress model; superconductor
integer :: tfc_model = 1
```

@stuartmuldrew Someone apparently though that a simple stress model *is* run when the coil is resistive:
```
    !  Ensure that if TF coils are non-superconducting,
    !  only simple stress calculations are performed
    if (itfsup == 0) tfc_model = 0
```
However, there doesn't seem to be any stress model attached.  
I have commented out all code that uses 'tfc_model'.

Test suite run successfully.

kj5248:
Closes #782.

Removed previously commented out sections of code relating to the tfc_model which is no longer used (unused in over 4 years). Removed mention of tfc_model in error relating to stellarators, and marked error 104 as obsolete (error to do with tfc_model).

### DIFF
--- a/process/utilities/errorlist.json
+++ b/process/utilities/errorlist.json
@@ -248,7 +248,7 @@
     {
       "no": 48,
       "level": 3,
-      "message": "LOADXC: Do not use thkcas as an iteration variable if tfc_model=0 or istell=1"
+      "message": "LOADXC: Do not use thkcas as an iteration variable if istell=1"
     },
     {
       "no": 49,
--- a/process/utilities/errorlist.json
+++ b/process/utilities/errorlist.json
@@ -527,8 +527,8 @@
     },
     {
       "no": 104,
-      "level": 3,
-      "message": "STRESSCL: Illegal value for tfc_model"
+      "level": 0,
+      "message": "OBSOLETE"
     },
     {
       "no": 105,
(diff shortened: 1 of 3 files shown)


timothy-nunn (approved):

timothy-nunn:
Michael, does this remove everything the original issue referred to

mkovari:
Yes, this seems to close out the original issue.  It seems that the Subroutine `sctfjalw` was removed a long time ago - this pull request just tidies up some code that was commented out anyway.

PR #3121, opened 2024-04-03
Fix Stellarator to allow it to run on open-source process

timothy-nunn:
## Description

Allows Stellarator to run by updating it to use the new physics interface post-conversion. 

Adds in the following calculation for VV stress to enable constraint 65:
```python
sctfcoil_module.vv_stress_quench = (
    f_vv_actual
    * 1e6
    * ((build_variables.d_vv_in + build_variables.d_vv_out) / 2)
)
```

Modifies the `SingleRun` initialisation order slightly so that the correct cost model is passed to Stellarator/IFE

### DIFF
--- a/process/main.py
+++ b/process/main.py
@@ -347,6 +347,7 @@ class SingleRun:
         self.validate_input()
         self.init_module_vars()
         self.set_filenames()
+        self.initialise()
         self.models = Models()
         self.solver = solver
 
@@ -355,8 +356,6 @@ class SingleRun:
 
         This is separate from init to allow model instances to be modified before a run.
         """
-        self.set_filenames()
-        self.initialise()
         self.validate_user_model()
         self.run_tests()
         self.call_solver()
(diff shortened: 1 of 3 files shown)


ym1906:
I was looking into some of the errors that are output after a single run. I think the order the models are called needs to be looked at to see if it makes sense.

Additionally, we assume in the Stellarator model that the plasma current is 0. Stellarators do have a lower plasma current than Tokamaks, but it may not necessarily be exactly 0.  However, this 0 plasma current is being used in some calculations which is throwing up divide by 0 errors. We should therefore consider how this is manifest in the code.

timothy-nunn:
The plasma current being 0 only seems to throw errors on output, so do we need to consider what we call for output when running stellarator because if stuff is diving by 0 the output probably doesn't make sense for a stellarator ?

ym1906:
Yes, looking into it further this parameter nu_star (normalized collisionality) may be a property of interest but the calculation is just lifted directly from tokamak output and it doesn't work so we might as well remove it for now.

ym1906:
Further to this, a lot of the post-processing output errors are cleared up when you use Jorrit's input file. The Felix-Stuart one is missing a few things which lead to these errors.

ym1906 (approved):
It looks good to me!

Issue #392, opened 2016-06-02
Olivier Sauter's formulae including negative triangularity


jonmaddock:
In GitLab by @hlux on Jun 2, 2016, 10:42

Please include Oliviers new formula in PROCESS, so that we can run comparisons with M. Kikuchi's suggested reactor version.

mkovari:
The reactor parameters for Kikuchi's negative triangularity configuration, using Sauter's new formulae valid for negative delta, are given on pp 40-41 of the link above.

mkovari:
`qpsi     : input real :  plasma edge safety factor (= q-bar for icurr=2)`
In the calling function we find `physics_variables.q`.  This is passed to `culcur` where it is named `qpsi` for some reason.  This is consistent with the documentation:

q | Input | real | 3.0 | Safety factor 'near' plasma edge (iteration variable 18) equal to q95  (unless icurr=2 (ST current scaling), in which case q = mean edge safety factor qbar)
-- | -- | -- | -- | --

- [ ] Update docstring in `culcur`.
- [ ] Rename `qpsi` in `culcur` as `q95`.
- [ ] Eliminate `curhat` since it's not necessary.
- [ ] Test for a range of values of triangularity $\delta$.

mkovari:
When the triangularity is negative, the vacuum vessel may clash with the TF coil, not surprisingly:
![image](https://github.com/ukaea/PROCESS/assets/9718416/2ecf5f15-f9fc-491f-aa4e-a8c63154086b)
I have not attempted to fix this.

mkovari:
Unfortunately, Kikuchi's presentation doesn't clearly give the values of elongation or triangularity, so I can't use it for benchmarking.  I won't be able to do a quantitative check on the geometric formulae.

I will try a qualitative check, using two runs with triangularity of opposite sign.  

- The poloidal perimeter is the same, because the formula contains only $\delta^2$.  
- The cross-section is also the same, because the formula doesn't include $\delta$ at all.
- The surface area and volume are bigger for negative $\delta$, because the mean radius is bigger - the plasma is further from the axis on average.

This seems OK.

  |   | Negative | Positive |  
-- | -- | -- | -- | --
Major radius (m) | (rmajor) | 8 | 8 | ITV
Minor radius (m) | (rminor) | 2.667 | 2.667 | OP
Aspect ratio | (aspect) | 3 | 3 |  
Elongation, X-point (input value used) | (kappa) | 1.85 | 1.85 | IP
[...]

PR #3128, opened 2024-04-05
#392 Sauter's formulae for plasma current and geometric parameters.

mkovari:
Minor refactor of `culcur` in `physics.py`, as described in #392:
Update docstring.
Rename qpsi in culcur as q95.
Eliminate curhat since it's not necessary.
Test for a range of values of triangularity.

### DIFF
--- a/process/plasma_geometry.py
+++ b/process/plasma_geometry.py
@@ -554,7 +554,7 @@ class PlasmaGeom:
 
         return xsect0
 
-    def Sauter_geometry(self, a, r0, kap, tri):
+    def sauter_geometry(self, a, r0, kap, tri):
         """
         Plasma geometry based on equations (36) in O. Sauter, Fusion Engineering and Design 112 (2016) 633–645
         'Geometric formulas for system codes including the effect of negative triangularity'
--- a/process/plasma_geometry.py
+++ b/process/plasma_geometry.py
@@ -222,24 +222,8 @@ class PlasmaGeom:
             physics_variables.kappa,
             physics_variables.triang,
         )
-
-        #  Poloidal perimeter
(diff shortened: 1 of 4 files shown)


mkovari:
`E       TypeError: culcur() got an unexpected keyword argument 'qpsi'`
I have renamed `qpsi` in culcur as `q95`.
@timothy-nunn

mkovari:
See detailed comments in the issue #392.

process/plasma_geometry.py
```diff
@@ -534,3 +553,40 @@ def xsect0(self, a, kap, tri):
         xsect0 = xlo**2 * (thetao - cto * sto) + xli**2 * (thetai - cti * sti)
 
         return xsect0
+
+    def Sauter_geometry(self, a, r0, kap, tri):
```
timothy-nunn, line 557: Even though its a persons name, I think `Sauter` should be lower case

process/plasma_geometry.py
```diff
@@ -249,10 +233,45 @@ def geomty(self):
             thetao,
         )
         physics_variables.sareao = xso
-        physics_variables.sarea = xsi + xso
 
-        #  Cross-sectional area
-        physics_variables.xarea = self.xsecta(xi, thetai, xo, thetao)
+        # icurr = 8 specifies use of the Sauter geometry as well as plasma current.
```
timothy-nunn, line 237: Will it ever make sense to use Sauter geometry without using the Sauter plasma current?

timothy-nunn (changes requested):
One minor change for code quality, one question to ensure this change is utilised to its fullest potential

process/plasma_geometry.py
timothy-nunn: Ok. If you make this into a new issue we can discuss it there

PR #3130, opened 2024-04-08
Use full double precision in MFILE output

jonmaddock:
Use full precision of double float in MFILE output; required for precise solution vectors and reliability analysis studies.

### DIFF
--- a/source/fortran/output.f90
+++ b/source/fortran/output.f90
@@ -485,7 +485,11 @@ contains
     call underscore(dum20)
     write(mfile,10) dum72, dum20, value, flag
 
-10  format(1x,a,t75,a,t110,1pe11.4," ",a,t10)
+  ! MFILE.DAT format
+  ! Machine epsilon for double ~2.22e-16, hence require 17 sig figs in significand
+  ! for full precision of a double float
+10  format(1x,a,t75,a,t110,ES23.16e2," ",a,t10)
+  ! OUT.DAT format
 20  format(1x,a,t75,a,t100,1pe10.3, t112, a)
 
   end subroutine ovarre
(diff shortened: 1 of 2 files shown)


timothy-nunn (approved):
@jonmaddock I'll approve this but won't merge it since you have not yet requested review. Feel free to merge it.

jonmaddock:
Hi Tim, I've made some changes regarding the regression tolerance that I think are worth you reviewing. Because we've now got full precision in the MFILE.DAT output, there's some floating-point discrepancy between running the regression test locally and on the CI. So the reference (run on the CI (tracker)) would disagree with local (explicitly-requested 0% tolerance) running due to full precision assertions. I've decided to go back to pytest's `approx()` for this, as it contains clear and proper handling for values close to 0 by using a combination of relative and absolute checks with sensible defaults.

timothy-nunn (changes requested):
This new method does not appear to respect the `--reg-tolerance` as exactly as it should. 

E.g. I made changes to Cost c25 in the source by multiplying the output value to check the test suite respected the default 5% tolerance. The old code would error as soon as I multiplied by anything ober `1.05`, e.g. `1.051`.

However, the new code using `approx` does not fail when multiplying by `1.051`, `1.052`... it fails at `1.053`. 

(note I did these tests using both a remote reference MFile and by setting a local one to be the reference: `reference_mfile_location = INPUT_FILES_FOLDER / "large_tokamak.MFILE.DAT"`).

jonmaddock:
Thanks for testing this! I think the reason was due to me using `pytest.approx()` on the observed (new) value, rather than the reference (old) one; the relative tolerance was being calculated relative to the observed value, not the reference. I've swapped the two around and convinced myself with an example, but you may want to check again to be sure.

timothy-nunn (approved):

PR #3135, opened 2024-04-12
Sort optimisation parameter numbers

jonmaddock:
Ensures consistent order and makes output comparison easier.

### DIFF
--- a/process/main.py
+++ b/process/main.py
@@ -427,7 +427,14 @@ class SingleRun:
     def initialise():
         """Run the init module to call all initialisation routines."""
         initialise_imprad()
+        # Reads in input file
         fortran.init_module.init()
+
+        # Order optimisation parameters (arbitrary order in input file)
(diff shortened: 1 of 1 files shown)


jonmaddock:
As the optimisation parameters are now sorted immediately after reading in from the input file, this causes the solution vector in the mfile to be re-ordered. The (re-ordered) solution vector is very similar, but not the same; this is likely to be to do with the non-idempotence effect previously observed when running the models multiple times or re-ordering the optimisation parameters. This can be reduced or eliminated by re-running the models until the outputs converge (https://github.com/ukaea/PROCESS-usage/issues/12). I think this effect is separate to this issue, and therefore although the resulting MFILE is (slightly) different, it is no less valid than the current regression reference.

timothy-nunn:
Note that comparing `itvar*` and `xcm*` values do not make sense due to the re-ordering of the inputs.

timothy-nunn (approved):

PR #3155, opened 2024-04-24
3148 kallenback model detailed in docs despite not being in process

j-a-foster:
## Description

Removes Kallenbach model mentions from docs as model has been removed from PROCESS.

## Checklist


timothy-nunn (changes requested):
Ill be honest, we should just remove rather than commenting. If it comes back, it will be a while and the documentation is in the history!

timothy-nunn (approved):

PR #3163, opened 2024-04-25
Update relative links in process docs

timothy-nunn:
## Description
Corrects relative image links.

**This PR does not include testing to avoid this happening in future**
The only way I can see to (easily) test this would be to check `mkdocs build --strict` does not fail. However, by the nature of our docs, some warnings will *always* occur. E.g.
```
WARNING -  [git-revision-date-localized-plugin] '/Users/timothynunn/process/documentation/proc-pages/io/vardes.md' has no git logs, using current
           timestamp
WARNING -  [git-revision-date-localized-plugin] '/Users/timothynunn/process/documentation/proc-pages/io/vardes.md' has no git logs, using current
           timestamp
```

@j-a-foster I know this doesn't solve your issue, but I don't think its possible without significant time investment that is likely not worth it.

### DIFF
--- a/documentation/proc-pages/eng-models/central-solenoid.md
+++ b/documentation/proc-pages/eng-models/central-solenoid.md
@@ -85,7 +85,7 @@ in the case of the CS coils).
     11. A safety factor of 2 is used for the crack size.   
 
 <figure markdown>
-![CS defect](../../images/conductor_coordinates.png){ width="100%"}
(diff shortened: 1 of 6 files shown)


j-a-foster (approved):
Changes to links has removed warnings when docs are generated. Remaining warnings are from dynamically generated docs. Ready for merge.

PR #3173, opened 2024-05-02
Added casths_fraction and casths as iteration variables

chris-ashe:
## Description

`casths_fraction` and `casths` have now been implemented as iteration variables

## Checklist

### DIFF
--- a/source/fortran/iteration_variables.f90
+++ b/source/fortran/iteration_variables.f90
@@ -3805,7 +3805,7 @@ contains
   end subroutine set_itv_170
 
   subroutine init_itv_171
-    !! <LI> (171) casths: TF side case fraction of toroidal thickness
+    !! <LI> (171) casths_fraction: TF side case fraction of toroidal thickness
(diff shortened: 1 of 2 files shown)


mkovari:
It would be nice at some point to add the relevant reference dimension (total coil thickness at the inner radius of the WP) to Figure 1 of the  [docs](https://ukaea.github.io/PROCESS/eng-models/tf-coil/#tf-coil-inboard-mid-plane-geometry).

mkovari:
You could doo with a check to make sure that the two new iteration variables are not used at the same time.

chris-ashe:
> You could doo with a check to make sure that the two new iteration variables are not used at the same time.

I have checked and with both variables iterating there is no difference to the number of steps to convergence, as default PROCESS will still override `casths` if `casths_fraction` is defined in the input. This is explained in the docs for the user

timothy-nunn (approved):
Implementation correct. Others seem happy with the change also.

PR #3176, opened 2024-05-09
Reset working directory after each test

timothy-nunn:
Running `pytest` on main was, again, causing the regression tests to fail with error `128`.

It seems that since writing the fix in #3162, another test has been introduced that unsafely changes the working directory. 
I have moved the fix in #3162 to a fixture that will run after each test to avoid this game of cat and mouse.

NOTE: this fixture can (and will) be used to identify which tests are doing naughty directory changes.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,6 +3,7 @@
 Defines fixtures that will be shared across all test modules.
 """
 
+import os
 import pytest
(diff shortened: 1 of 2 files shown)


clmould (approved):

jonmaddock:
Thanks Clair!

PR #3177, opened 2024-05-09
277 ndscan tools library removal

ym1906:
Removed legacy ndscan tools for papercut issue: https://github.com/users/je-cook/projects/4/views/4?sliceBy%5Bvalue%5D=ym1906&filterQuery=&pane=issue&itemId=53390543


timothy-nunn (approved):
`NDScanConfig` class never used, this PR removes it and all functions only used within it.

PR #3179, opened 2024-05-15
Do not compare MFile error variables

timothy-nunn:
## Description

When comparing MFiles using `mfile_comparison.py`, do not compare the `error_status` or `error_id` in the MFile.

### DIFF
--- a/process/io/mfile_comparison.py
+++ b/process/io/mfile_comparison.py
@@ -1,7 +1,9 @@
 #!/usr/bin/env python
 """
 
-    Python tool for comparing MFILE and outputting differences
+    Python tool for comparing MFILE and outputting differences.
+    The tool does not work for MFiles that are not the result of
+    a full PROCESS run (ie if an error or exception occured).
 
     James Morris
     14/04/15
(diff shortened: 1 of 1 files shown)


jonmaddock (commented):
I'm not sure this is quite what we want. The original issue was from when this utility was used in the CI system, which is no longer the case. It was to fail the CI if a level 3 error was reported, not to pass if the error was still a level 3. I think we want to keep error differences being reported in the mfile differences: that's a pretty key difference!

Sorry if I've got the wrong end of the stick. Code is fine, obviously!

timothy-nunn:
Hi @jonmaddock,

Sorry -- I've assumed this issue is to disable error comparison because it does not make sense and was causing an error when comparing seemingly identical MFiles. The issue was more asking us to disable comparison of error3 mfiles entirely.

jonmaddock:
Ok, so now if a regression test runs and results in a level 3 error (i.e. an error is encountered before a solution can be found), the mfile comparison won't go ahead, and the regression test for this input file only will fail due to the exception you're now raising? If that's the case, great!

jonmaddock (commented):
Ok, so now if a regression test runs and results in a level 3 error (i.e. an error is encountered before a solution can be found), the mfile comparison won't go ahead, and the regression test for this input file only will fail due to the exception you're now raising? If that's the case, great!

timothy-nunn:
Hi @jonmaddock,

Yes, that is the case. A `RuntimeError` occurs if any of the MFiles are not the result of converged PROCESS run (ie `error_status==3`)

jonmaddock (approved):
Great, that's clear.

PR #3181, opened 2024-05-16
Dont track unfound vars

timothy-nunn:
## Description

Do not track variables that cannot be found in the MFile


j-a-foster (approved):
This works pretty much as intended on my repo. The file names are still out of the keys of the tracker plots but I think that's just because the file names are too long.

PR #3182, opened 2024-05-16
Create new regression tests

timothy-nunn:
## Description

Adds regression tests for Stellarator, Spherical Tokamak, Large Tokamak (no f-values) and Large Tokamak solution (once through).

With thanks to @ym1906 and @chris-ashe!


jonmaddock (approved):
Great to get some more regression tests in!

PR #3197, opened 2024-05-23
Get ST to run and converge

timothy-nunn:
## Description

Fixes a deprecated variable in ST file and moves initial point slightly to allow the file to converge


chris-ashe (approved):
Initial values changed to enable smooth convergence

Issue #1114, opened 2020-09-18
Jc Nomenclature


jonmaddock:
In GitLab by @schislet on Sep 18, 2020, 18:05

Associated branch https://git.ccfe.ac.uk/process/process/-/tree/issue-1113-jc-nomenclature

## Summary

![image](https://git.ccfe.ac.uk/process/process/uploads/7a74107ab260d1bbc6f0729c07b2fcbb/image.png)

The critical current density of the superconducting cable ignores the fraction of the cable taken up by the coolant channel - this is taken into account only in the calculation of the cable critical current, rather than its critical current density. For clarity I would recommend that the cooling fraction is taken account in the critical current density calculation step by introducing an additional local variable j_crit_cable, the critical current density of the cable e.g.:   
~~~~~~~~~~~~~~~
call wstsc(thelium,bmax,strain,bc20m,tc0m,jcritsc,bcrit,tcrit)  
            jcritstr = jcritsc * (1.0D0-fcu)  
            ! Critical current density in cable  
            j_crit_cable = jcritstr* fcond  
            !  Critical current in cable  
            icrit = j_crit_cable * acs   
~~~~~~~~~~~~~~~~~
Also, the copper fraction of the superconducting strands (in the form of the surrounding copper matrix in superconducting strands) has been combined with the copper fraction within the cable (in the form of additional copper strands in Nb3Sn cables or REBCO tape support). This is not a computational problem, but is not clearly presented in the source code or documentation. For the LTS strands, the subroutines in superconductors.f90 output jc_non-copper (from the attached figure), but the GL_REBCO model outputs jc_total-tape. This is **not** clear from the descriptions in superconductors.f90. I shall update them.  

For REBCO (and Bi2212) fcu = copper fraction of conductor NOT including the copper layers in the REBCO tape itself, but for LTS fcu = copper fraction of the conductor INCLUDING the copper in individual strands. I have made this explicit when the GL_REBCO model is called:
~~~~~~~~~~~~~
call GL_REBCO(thelium,bmax,strain,bc20m,tc0m,jcritsc,bcrit,tcrit)  
            ! For this model fcu = NON TAPE copper fraction. GL_REBCO gives the   
            ! tape critical current inclusive of the tape's copper/buffer layers etc.   
            jcritstr = jcritsc * (1.0D0-fcu)  
~~~~~~~~~~~~~

For reference:

[...]

PR #3205, opened 2024-05-30
Papercut 1114 jc nomenclature

geograham:
## Description

Addresses Papercut #1114 

## Checklist


process/io/obsolete_vars.py
```diff
@@ -104,7 +104,10 @@
     "bcylth": None,
     "itfmod": None,
     "jcrit_model": None,
-    "jcritsc": None,
+    "jcritsc": "j_crit_sc",
+    "jcritstr": "j_crit_cable",
+    "jstrandoh_bop": "jcableoh_bop",
+    "jstrandoh_eof": "jcableoh_eof",
```
timothy-nunn, line 110: These variables do not need to be added here because they are not input variables.

timothy-nunn (changes requested):
Happy with the code changes. I'll ask for comments from @ajpearcey @chris-ashe @stuartmuldrew about impact of renaming output variables. While its technically backwards incompatible, it might be _ok_ for these variables. 

I can see arguments on both sides, especially for the `strand` -> `cable` variables where the variable name does not match the quantity. Is that worth backwards incompatibility?

cc @jonmaddock

process/io/obsolete_vars.py
geograham: Thanks, have removed.

timothy-nunn (approved):
Spoke with @chris-ashe who doesn't believe these changes will affect downstream scripts. Bluemira does not read these entries in our MFile.

While this is technically a breaking API change, I do not believe it will be in practice. Given the lack of comments and the fact these output names are misleading, I am happy to approve the change.

PR #3209, opened 2024-05-31
Removed tfsai and tfsao from sctfcoil due to obsolete var in calcs

j-a-foster:
## Description

I removed `tfsai` and `tfsao` from `sctfcoil.py` as they were using the variable `hr1` which wasn't calculated or an input and just set to 0. A power calculation that used `tfsai` was then replaced with `tfcryoarea` based on comments in the code (which have now been removed).

## Checklist

### DIFF
--- a/process/power.py
+++ b/process/power.py
@@ -789,7 +789,7 @@ class Power:
             # heat_transport_variables.helpow calculation
             heat_transport_variables.helpow = self.cryo(
                 tfcoil_variables.i_tf_sup,
-                tfcoil_variables.tfsai,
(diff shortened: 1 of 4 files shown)


j-a-foster:
The changes in the regression tests appear to make sense to me but would like a second opinion.

mkovari (approved):

mkovari:
Since the cryogenic model is pretty rubbish anyway, I don't think this deserves too much brain power.  In any case it's got to be better (?) to replace a variable that was never set by one that has a value.

Issue #283, opened 2015-05-07
Bremsstrahlungradiation to include Gaunt factor (low priority)


jonmaddock:
In GitLab by @hlux on May 7, 2015, 09:37

Thomas Puetterich suggested to update our Bremsstrahlungs calculation using a prescription for the Gaunt-factor as he thinks it will make a significant difference. This is currently low priority as the only two points affected by this are our output of the Bremsstrahlung power and the high T extrapolation (T_e > 100 keV) of the ADAS data. For the rest of the radiation treatment we are using the ADAS data, which is taking all quantum-mechanical and other effects into account.

ajpearcey:
If the Bremsstrahlung radiation is incorrect I suggest we remove it. See issue #1231.

PR #3210, opened 2024-06-03
Remove bremsstrahlung and line radiation calculations

### LINKED ISSUE #1231: Calculation of bremsstrahlung
In GitLab by @mkovari on Jan 14, 2021, 14:43

Issue #1230 showed that `function pbremden` (Bremsstrahlung radiation density) takes 32% of the total execution time (includes calls from that function).   
image  
*As far as I can see*, the only purpose of this function is to be able to divide the total radiation into line radiation and bremsstrahlung so as to be able to output these quantities.  I think that only the total radiation is actually used in the remainder of the PROCESS calculation.  The only line I can find where the bremsstrahlung is used is
```
pline = pimp - pbrem
```
where
```
!! pimp  : output real : total impurity radiation density (W/m3)
!! pbrem : output real : Bremsstrahlung radiation density (W/m3)
!! pline : output real : other radiation density (W/m3)
```

If this is correct, then we have some simple options for speeding up the code:  
EITHER Remove this routine entirely and remove the separate output of line radiation and bremsstrahlung,   
OR calculate the bremsstrahlung only on the function call that occurs after the last VMCON iteration when the OUT.DAT will be written.

Detailed options:
1.  Replace the function `zav_of_te` by something much quicker.
2.  Reduce the number of steps in the profile integration: `npts = 200` in `subroutine imprad`.

ajpearcey:
This closes #1231 and #283


ajpearcey:
This is now only failing the regression tests and is ready for review

mkovari:
I have run the large-tokamak on this branch and compared it to the result using `main`, with significant differences.  I think this is just because `main` has changed.  Would you be able to merge `main` into your branch and push it back onto GitHib so they can be compared?
@ajpearcey

ajpearcey:
This code removal causes a change in the solution vector. This is because of check for a model inconsistency that pimp was not smaller than pbrem, which was being found. Now after removing pbrem calculation this can never be true.  

image

mkovari (approved):

PR #3231, opened 2024-06-21
Make objective function plot optional

jonmaddock:
Allows plotting solutions with different objective functions.


timothy-nunn (changes requested):
Hi Jon,

When running `plot_mfile_solutions` with `rmse=True` and `plot_obj_func=False` I get an `UnboundLocalError`. Could you either fix this or raise a `ValueError` if this combination of arguments is illegal.

Code otherwise looks fine and works when `rmse=False`

PR #3240, opened 2024-07-02
Add additional inequality satisfaction convergence criterion

jonmaddock:
Ensure that inequality constraints are satisfied at the solution point.

This requires https://github.com/ukaea/PyVMCON/pull/16 to be merged first, to actually call any additional convergence criterion.

This change will only affect optimising regression tests without f-values (currently only `large_tokamak_nof.IN.DAT` and `stellarator.IN.DAT`.) However, the large tokamak case already has satisfied inequality constraints, so it doesn't change. The stellarator test takes over 400 iterations running locally, so I haven't bothered running it. Before this is merged, I'll find current evidence of this actually changing the solution to having all-satisfied inequality constraints.

### DIFF
--- a/process/solver.py
+++ b/process/solver.py
@@ -160,7 +160,6 @@ class Vmcon(_Solver):
     def solve(self) -> int:
         """Optimise using new VMCON.
 
-        :raises NotImplementedError: not currently implemented
         :return: solver error code
         :rtype: int
         """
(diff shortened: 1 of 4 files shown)


jonmaddock:
After checking again on an updated `main`, the stellarator regression test did converge after 18 iterations, but it had 2 violated inequality constraints. With this additional convergence criterion, it now does not converge.

I haven't been able to find a current example where this requirement (satisfaction of inequality constraints) results in a different, fully satisfied, converging solution, although I have experienced this with the large tokamak regression test in the past. Currently the large tokamak regression test converges with satisfied inequality constraints without this additional criterion anyway.

I believe this should be merged anyway, as a solution isn't feasible without this criterion being met.

timothy-nunn (approved):
LGTM. Please make an issue regarding re-enabling Stellarator regression test.

jonmaddock:
Thanks Tim: added #3245 for the stellarator regression test re-introduction.

PR #3242, opened 2024-07-03
Removed confusing comment about double null, added more descriptive c…

ym1906:
The calculation of hmax in the vertical build was a bit confusing as there was/is a comment that implies it's only valid for double null builds.

In general we assume TF coils to be symmetrical, so in fact this calculation is fine for both single and double null.

I have clarified this in the comment of the code.


ajpearcey (commented):
LGTM

ajpearcey (approved):

PR #3244, opened 2024-07-04
3243 remove gamfun and replace it with the analytical solution from scipy

chris-ashe:
## Description

The use of `gamfun()` from `maths_library.f90` within `plasma_profiles.py` for the temperature profile has been removed and replaced with `scipy.special.gamma`

## Checklist


j-a-foster (commented):
All looks good to me. I'm assuming where you've removed `maths_library` that it's only imported to call `gamfun`?

chris-ashe:
@j-a-foster  Have reverted the commit as it was causing problems with `init.f90` and their is still some function used from `maths_functions.f90`

tests/unit/test_plasma_profiles.py
```diff
@@ -246,15 +246,15 @@ class PlasmaProfilesParam(NamedTuple):
...
-            expected_ten=14.521871327399182,
-            expected_tin=14.521871327399182,
+            expected_ten=14.52233022043558,
+            expected_tin=14.52233022043558,
             expected_alphap=2.4500000000000002,
-            expected_te0=27.369013322953624,
-            expected_p0=868071.46874220832,
-            expected_pcoef=1.1110842637642833,
+            expected_te0=27.370104119511087,
+            expected_p0=868106.0658743214,
```
timothy-nunn, line 253: This change is quite big. Any idea?

tests/unit/test_plasma_profiles.py
```diff
@@ -291,15 +291,15 @@ class PlasmaProfilesParam(NamedTuple):
...
```
timothy-nunn, line 298: Again, why so much bigger than the others

timothy-nunn (changes requested):

tests/unit/test_plasma_profiles.py
chris-ashe: The calculation is several gamma terms multiplied together so the divergence in result is compounded at the output

tests/unit/test_plasma_profiles.py
chris-ashe: ^ see above

timothy-nunn (changes requested):
Line 52 (`integrate.simpson`) needs to become `sp.integrate...`

PR #3251, opened 2024-07-11
Plot-proc fixes in poloidal cross section for first wall and blanket

kj5248:
## Description

Changed the colour of the first wall fill for double null case to use first wall colour not the blanket colour. Also added small if statement to not plot the inboard blanket if it has 0 thickness for the double null case.

First image shows the poloidal cross section before any changes. The second shows the present of inner blanket when there shouldn't be overlapping with the first wall. Final image is post changes.

Before
inner_blanket
After

### DIFF
--- a/process/io/plot_proc.py
+++ b/process/io/plot_proc.py
@@ -1326,17 +1326,21 @@ def plot_blanket(axis, mfile_data, scan, colour_scheme) -> None:
         )
         # Plot blanket
         axis.plot(bg_double_null.rs[0], bg_double_null.zs[0], color="black", lw=thin)
-        axis.plot(bg_double_null.rs[1], bg_double_null.zs[1], color="black", lw=thin)
         axis.fill(
             bg_double_null.rs[0],
             bg_double_null.zs[0],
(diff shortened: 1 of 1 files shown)


process/io/plot_proc.py
```diff
@@ -1326,17 +1326,21 @@ def plot_blanket(axis, mfile_data, scan, colour_scheme) -> None:
...
             bg_double_null.rs[0],
             bg_double_null.zs[0],
             color=BLANKET_COLOUR[colour_scheme - 1],
         )
-        axis.fill(
-            bg_double_null.rs[1],
-            bg_double_null.zs[1],
-            color=BLANKET_COLOUR[colour_scheme - 1],
-        )
+        if blnkith > 0:
```
chris-ashe, line 1334: Should this not be 0.0 as `blnkith` is a `float`

kj5248: Well spotted, done

chris-ashe (approved):
Changes have been implemented and the code now prevents the inboard blanket being plotted when not present

PR #3265, opened 2024-08-02
Calculation of Tresca stress in TF winding pack

### LINKED ISSUE #580: Calculation of Tresca stress in TF winding pack
In GitLab by @mkovari on Nov 9, 2017, 08:45

Section 2.4.3 of the PROCESS engineering paper explains that 
>Because the structural material inside the winding pack is assumed to take the form of radial and tangential webs (Fig. 7), the limiting von Mises stress in this zone is given by the larger of the two values:  
image

I think we should adopt the same method when calculating the two Tresca stresses: 
1. Calculate the radial stress in the radial webs, and combine with the vertical stress;
2. Calculate the tangential stress in the tangential webs, and combine with the vertical stress.

We need to think about how these are derived from the single set of smeared stresses.

ajpearcey:
Closes #580

### DIFF
--- a/process/sctfcoil.py
+++ b/process/sctfcoil.py
@@ -4254,11 +4254,8 @@ class Sctfcoil:
         # Tresca / Von Mises yield criteria calculations
         # -----------------------------
         # Array equation
-        sig_tf_tresca_tmp1 = numpy.maximum(
-            numpy.absolute(sig_tf_r), numpy.absolute(sig_tf_r - sig_tf_z)
-        )
         sig_tf_tresca = numpy.maximum(
-            sig_tf_tresca_tmp1, numpy.absolute(sig_tf_z - sig_tf_t)
+            numpy.absolute(sig_tf_r - sig_tf_z), numpy.absolute(sig_tf_z - sig_tf_t)
         )
 
         # Array equation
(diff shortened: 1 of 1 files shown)


process/sctfcoil.py
```diff
@@ -4255,7 +4255,7 @@ def stresscl(
         # -----------------------------
         # Array equation
         sig_tf_tresca_tmp1 = numpy.maximum(
-            numpy.absolute(sig_tf_r - sig_tf_t), numpy.absolute(sig_tf_r - sig_tf_z)
+            numpy.absolute(sig_tf_r), numpy.absolute(sig_tf_r - sig_tf_z)
         )
         sig_tf_tresca = numpy.maximum(
             sig_tf_tresca_tmp1, numpy.absolute(sig_tf_z - sig_tf_t)
```
mkovari, line 4261: I am not sure why you have included $|\sigma_r|$.  It shouldn't make any difference because $\sigma_z$ should always be negative, but I don't think you need it.

So I think all you need is one line
```python
 sig_tf_tresca_tmp1 = numpy.maximum(
            numpy.absolute(sig_tf_z - sig_tf_t), numpy.absolute(sig_tf_r - sig_tf_z)
)
```

mkovari (approved):
Any changes to the output?

PR #3269, opened 2024-08-07
Current ramp-up #489

### LINKED ISSUE #489: Current ramp-up time
In GitLab by @mkovari on Mar 2, 2017, 11:08

The consumption of flux during start-up and current ramp is given by sum of the resistive flux consumption during start-up and current ramp and the flux consumption due to the self-inductance of the plasma during current ramp-up. The calculation of resistive flux consumption is based on the fact that the ramp-up takes of the order of the resistive current penetration time. The flux consumption is therefore dependent on the empirical Ejima constant, defined by the user (default value 0.4).

_However_, we don't actually calculate the ramp-up time in this way!  There are a couple of limitations on the ramp-up time:

+ Constraint 41: Plasma current ramp-up time lower limit   (The minimum inductive current ramp-up time depends on the voltage available from the CS power supply and the inductive coupling between the CS and the plasma).  
+ Constraint 66: Limit on rate of change of energy in poloidal field. 
+ In addition the maximum rate of change of plasma current  is a function of the plasma current:
` ipdot = 0.0455D0*plascur`

When the plasma reaches its burn temperature the resistive diffusion time is very long, and it is this time that is implied in the flux calculation using the Ejima coefficient:  
```
Resistive diffusion time (s)         (res_time)                3.170E+03  OP 
```
The current ramp actually takes place at much lower temperatures, and does not take this long.

In practice this means that the calculation of the required flux swing is probably OK as long as the current ramp is set to a value less than the resistive diffusion time `res_time`.

Comments @rkemp @jmorris-uk @fmavig

mkovari:
## Checklist

- [x] I have justified any large differences in the regression tests caused by this pull request in the issue #489. 
- [x] I have added new tests where appropriate for the changes I have made - None
- [x] Unit test updated.
- [x] If I have made documentation changes, I have checked they render correctly - None
- [x] I have added NOT yet added documentation for my change.

### DIFF
--- a/process/physics.py
+++ b/process/physics.py
@@ -205,8 +205,6 @@ def rether(alphan, alphat, dene, dlamie, te, ti, zeffai):
 
     return conie * (ti - te) / (te**1.5)
 
-
-# @nb.jit(nopython=True, cache=True)
 def vscalc(
     csawth,
     eps,
--- a/process/physics.py
+++ b/process/physics.py
@@ -206,7 +206,7 @@ def rether(alphan, alphat, dene, dlamie, te, ti, zeffai):
     return conie * (ti - te) / (te**1.5)
 
 
-@nb.jit(nopython=True, cache=True)
+# @nb.jit(nopython=True, cache=True)
 def vscalc(
     csawth,
     eps,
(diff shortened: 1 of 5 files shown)


process/physics.py
```diff
@@ -206,7 +206,7 @@ def rether(alphan, alphat, dene, dlamie, te, ti, zeffai):
     return conie * (ti - te) / (te**1.5)
 
 
-@nb.jit(nopython=True, cache=True)
+# @nb.jit(nopython=True, cache=True)
```
ajpearcey, line 209: Why is the jit decorator commented out? Should this be re-enabled?

ajpearcey (changes requested):

ajpearcey (approved):

mkovari:
Chris, could you remind me why you marked this PR as Blocked?

PR #3281, opened 2024-08-12
Don't sudo git commands in tracking

timothy-nunn:
Remove the need to `sudo` git commands in the CI. 

- **Why sudo?** I had accidentally set one of the `git` commands to run with `--system` and not `--global`. Therefore, when we stopped running on our own container (as root) it became necessary to run the system level command as superuser (because the CI now runs with user `runner`). 
- **The problem with sudo?** It runs the command as root... a different user to the one the SSH client was setup for. Therefore, the superuser does not have the credentials to pull or push from the repo. Hence, the tracking job and regression test job fail with a variety of config missing or SSH credentials invalid errors.
- **The solution:** remove `sudo git` commands and just use `git`. This makes it necessary to replace `--system` with `--global`.

### DIFF
--- a/.github/workflows/process.yml
+++ b/.github/workflows/process.yml
@@ -123,7 +123,7 @@ jobs:
           pip install -e '.[test, examples]'
           pip install -r requirements_dev.txt
       - name: Allow git commands to be run
-        run: sudo git config --system --add safe.directory '*'
+        run: git config --global --add safe.directory '*'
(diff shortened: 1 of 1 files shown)


timothy-nunn:
Tested all actions running: https://github.com/timothy-nunn/PROCESS/actions/runs/10350468763

clmould (approved):

PR #3287, opened 2024-08-14
Added new argument to write json from mfile, added execution of argument to post processing

ym1906:
It is useful to have PROCESS dump the mfile to json for UQ work. I have added an argument to do this which is executed in postprocessing

### DIFF
--- a/process/main.py
+++ b/process/main.py
@@ -78,8 +78,6 @@ from process.caller import write_output_files
 from pathlib import Path
 import os
 import logging
-import numpy as np
-import json
 
 # For VaryRun
 from process.io.process_config import RunProcessConfig
--- a/process/main.py
+++ b/process/main.py
@@ -189,7 +189,7 @@ class Process:
             "-mj",
             "--mfilejson",
             action="store_true",
-            help="produce a json containing mfile data",
+            help="Produce a filled json named mfile.DAT.json in working dir.",
         )
 
         # If args is not None, then parse the supplied arguments. This is likely
--- a/process/main.py
+++ b/process/main.py
@@ -189,7 +189,7 @@ class Process:
             "-mj",
             "--mfilejson",
             action="store_true",
-            help="Produce a filled json named mfile.DAT.json in working dir.",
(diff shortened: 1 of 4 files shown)


process/main.py
```diff
@@ -209,14 +216,20 @@ def post_process(self):
         # run, for example.
         if self.args.plot:
             # Check mfile exists, then plot
-            mfile = Path(self.args.mfile)
+            mfile_path = Path(self.args.mfile)
             mfile_str = str(mfile.resolve())
```
timothy-nunn, line 220:
```suggestion
            mfile_str = str(mfile_path.resolve())
```

process/main.py
```diff
@@ -184,6 +185,12 @@ def parse_args(self, args):
             default="MFILE.DAT",
             help="mfile for post-processing/plotting",
         )
+        parser.add_argument(
+            "-mj",
+            "--mfilejson",
+            action="store_true",
+            help="produce a json containing mfile data",
```
timothy-nunn, line 192: Update the help message slightly. Where does it write the JSON file, and with what name?

timothy-nunn (changes requested):
Please could you add an integration test for this (maybe with a once through input file?):
- Does running with this command produce a JSON file at the expected location with the expected name?
- Does the JSON file contain valid JSON?
- Does the JSON contain some outputs we would expect?

process/main.py
ym1906: parser.add_argument(
            "-mj",
            "--mfilejson",
            action="store_true",
            help="Produce a filled json from --mfile arg in working dir",
        )
Modified like this

process/main.py
ym1906: if self.args.plot:
            # Check mfile exists, then plot
            mfile_path = Path(self.args.mfile)
            mfile_str = str(mfile_path.resolve())

ym1906:
I addressed the changes, let me know if further improvements are needed.

timothy-nunn (approved):

Issue #3250, opened 2024-07-11
Fill not working as intended in plot_proc.py


kj5248:
## Describe the bug

In plot_proc matplotlib.pyplot.fill is frequently used in a way that it doesn't fill inside the shapes it is supposed to. This is because it has some linewidth which goes outside the shape.  See image below where on left is how it should fill and on the right how it currently does it.

![fill_example](https://github.com/ukaea/PROCESS/assets/115085094/471ad1dc-75d7-482e-b45d-83649259d75a)

This needs to be checked for the cases in plot_proc and whether it is an issue in these cases (and elsewhere?). Can be resolved by adding lw statement to fill:

```
axis.fill(
            xpoints,
           ypoints,
            color=colour,
            lw=thin
)
```

PR #3322, opened 2024-09-19
Plot proc fix for poloidal cross section

kj5248:
## Description

Made changes to poloidal cross section plotting. Changed component shape filling from default to a specified linewidth so that components aren't larger than intended and overlapping. This occurs sometimes for thin components as seen in issue #3250. Value was set to "0.01" instead of "thin" so that very thin components like the first wall don't disappear entirely.

ST Before:
st_before
ST After:
st_after
Tokamak Before:
tokamak_before
Tokamak After:
tokamak_after


chris-ashe (approved):
LGTM

PR #3324, opened 2024-09-20
Support array iteration variables when writing new indat

timothy-nunn:
## Description

* Modify `test_write_new_in_dat` to check `fimp(13)` assumes the value from the MFile in the new INDat (it initially fails)
* Use regex in `replace_iteration_variables` to identify array variables and use the `change_array` method instead of `add_parameter`.
* Cleanup some old comments

The modified test now passes, showing that the fix was successful


process/io/write_new_in_dat.py
```diff
@@ -106,9 +97,12 @@ def replace_iteration_variables(iteration_vars, in_data):
...
 
-    for name in iteration_vars.keys():
-        varname = name.lower()
-        in_data.add_parameter(varname, iteration_vars[name])
+    for variable_name, variable_value in iteration_vars.items():
+        
+        if (match := re.search(r"([a-zA-Z0-9_]+)\(([0-9]+)\)", variable_name)) is None:
+            in_data.add_parameter(variable_name.lower(), variable_value)
+        else:
+            in_data.change_array(match.group(1), int(match.group(2))-1,variable_value)
```
clmould, line 105: Can you add a space before `variable_value` please

tests/integration/test_write_new_in_dat.py
```diff
@@ -20,13 +20,16 @@ def test_write_new_in_dat(temp_data, mfile_name):
     # Get final value of te optimisation parameter
```
clmould, line 20: Can you add here that you are getting the final value of fimp as well as te

clmould (changes requested):

clmould (approved):
Looks good to me! :)

PR #3325, opened 2024-09-20
Fallback to minmax when plotting old solutions

timothy-nunn:
## Description

Uses the `minmax` output parameter when the newer `objf_name` parameter is not present (ie in old mfiles file)

When comparing more than 1 MFile, if any of them are missing `objf_name` it defaults to `minmax`.

Updated `tests/integration/data/large_tokamak_1_MFILE.DAT` to reflect this possibility.

### DIFF
--- a/process/io/plot_solutions.py
+++ b/process/io/plot_solutions.py
@@ -9,6 +9,8 @@ currently.
 """
 
 from process.io.mfile import MFile
+from process.fortran import numerics
+from process.utilities.f2py_string_patch import f2py_compatible_to_string
 from pathlib import Path
 import pandas as pd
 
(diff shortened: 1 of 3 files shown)


process/io/plot_solutions.py
```diff
@@ -342,12 +346,26 @@ def _plot_solutions(
     norm_objf_values_df = norm_objf_df.filter(regex=f"{NORM_OBJF_VALUE}|{TAG}")
 
     # Acquire objective function name(s), then check only one type is being plotted
-    objf_list = norm_objf_df[NORM_OBJF_NAME].unique()
+    if (
```
jonmaddock, line 349: Could you comment why this is required?

jonmaddock (changes requested):

jonmaddock (approved):

PR #3327, opened 2024-09-20
Increase lablxc length to 30

timothy-nunn:
Increase the string length of `lablxc`'s to 30 so that iteration variable names are not cut off when accessed from Fortran


clmould (approved):
Looks good to me!

PR #3330, opened 2024-09-24
3245 fix stellarator regression test

timothy-nunn:
Finds a new initial point for the Stellarator regression test. I had to relax some of the constraints and change the initial point using a vary run.

This initial point appears to be very sensitive so I suspect we will see this regression test failing in the future. 

[stellarator.MFILE.DAT.txt](https://github.com/user-attachments/files/17116831/stellarator.MFILE.DAT.txt)
[stellarator.OUT.DAT.txt](https://github.com/user-attachments/files/17116832/stellarator.OUT.DAT.txt)

### DIFF
--- a/tests/regression/input_files/stellarator.IN.DAT
+++ b/tests/regression/input_files/stellarator.IN.DAT
@@ -13,7 +13,7 @@ icc = 11                *Radial build (consistency equation)
 
 * Inequalities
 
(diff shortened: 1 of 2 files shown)


ym1906 (approved):
These are minor numerical changes and should not drastically change the design. Looks good to me.

PR #3339, opened 2024-10-02
Plot thermal shield in poloidal plot

timothy-nunn:
Plot the thermal shield in the poloidal plot in `plot_proc`. The way to plot the shield utilises the TF coil plotting routine to plot 3 TF coils shapes:
1. A tf coil shape with an inboard offset of `thshield_ib + tftsgap`, plot with the colour of the thermal shield.
2. A tf coil shape with an inboard offset of `tftsgap`, plot with a white colour.
3. A tf coil shape with an inboard offset of 0, plot with the TF coil colour.

When plot in this order, the shapes will overlap. It plots the thermal shield (with a white gap) a constant distance away from the inside of the TF coil, as shown (thermal shield is the thin grey line, this is reflected in the legend):

image

image

### DIFF
--- a/process/geometry/tfcoil_geometry.py
+++ b/process/geometry/tfcoil_geometry.py
@@ -16,6 +16,8 @@ def tfcoil_geometry_rectangular_shape(
     y4: float,
     y5: float,
     tfcth: float,
+    *,
+    offset_in: float = 0.0,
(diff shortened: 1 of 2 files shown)


chris-ashe (approved):
Discussed with @timothy-nunn. Changes are implemented correctly for both coil cases.

PR #3342, opened 2024-10-08
Correct comments in constraint_equations.f90

mkovari:
Only comments have been changed to correct copy-paste errors.

### DIFF
--- a/source/fortran/constraint_equations.f90
+++ b/source/fortran/constraint_equations.f90
@@ -1088,7 +1088,7 @@ contains
    end subroutine constraint_eqn_018
 
    subroutine constraint_eqn_019(tmp_cc, tmp_con, tmp_err, tmp_symbol, tmp_units)
-      !! Equation for MVA upper limit: resistive TF coil set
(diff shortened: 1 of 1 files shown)


source/fortran/constraint_equations.f90
```diff
@@ -1088,7 +1088,7 @@ subroutine constraint_eqn_018(tmp_cc, tmp_con, tmp_err, tmp_symbol, tmp_units)
    end subroutine constraint_eqn_018
 
    subroutine constraint_eqn_019(tmp_cc, tmp_con, tmp_err, tmp_symbol, tmp_units)
-      !! Equation for MVA upper limit
+      !! Equation for MVA upper limit: resistive TF coil set
```
timothy-nunn, line 1091:
```suggestion
      !! Equation for MVA (power) upper limit: resistive TF coil set
```

Or something similar to clarify what this is

timothy-nunn (changes requested):

timothy-nunn (approved):

PR #3346, opened 2024-10-10
Add IFE regression test

timothy-nunn:
Fixes bugs in the 1990 cost model for calculating IFE costs.

Initialises relevant IFE iteration variables 

Adds old IFE input file as regression test


stuartmuldrew (approved):
Input file looks the same as before and all info is fine to Open Access as it is based on published data.  There's no output file in this merge to see if the solution is the same.

PR #3351, opened 2024-10-17
Warn about non-idempotent values in MFILE

jonmaddock:
Warn, rather than raise exception, when MFILE variables are non-idempotent (still varying after 10 model evaluations). The variables and their values are put into a table in stdout.

### DIFF
--- a/process/caller.py
+++ b/process/caller.py
@@ -197,6 +197,7 @@ class Caller:
             # Values haven't all stabilised after 10 evaluations
             # Which variables are still changing?
             warnings.warn(
+                "\033[93m"
                 "Model evaluations at the current optimisation parameter vector "
                 "don't produce idempotent values in the final output."
             )
@@ -206,6 +207,7 @@ class Caller:
                     [[k, v[0], v[1]] for k, v in nonconverged_vars.items()],
                     headers=["Variable", "Previous value", "Current value"],
                 )
+                + "\033[0m"
             )
 
             # Close idempotence files, write final output file and mfile
--- a/process/caller.py
+++ b/process/caller.py
@@ -49,31 +49,6 @@ class Caller:
         else:
             return False
 
-    @staticmethod
-    def check_mfile_agreement(
-        previous: dict[str, float], current: dict[str, float]
-    ) -> dict[str, list[float, float]]:
-        """Compare previous and current mfiles for agreement within a tolerance.
-
-        :param previous: variables and values from previous models evaluation
-        :type previous: dict[str,float]
-        :param current: variables and values from current models evaluation
-        :type current: dict[str,float]
-        :return: non-converged values
-        :rtype: dict[str, list[float, float]
-        """
-        nonconverged_vars = {}
-        for var in previous.keys():
-            if np.allclose(
-                previous[var], current.get(var, np.nan), rtol=1.0e-6, equal_nan=True
-            ):
-                continue
-            else:
-                # Value has changed between previous and current mfiles
-                nonconverged_vars[var] = [previous[var], current.get(var, np.nan)]
-
-        return nonconverged_vars
-
     def call_models(self, xc: np.ndarray, m: int) -> Tuple[float, np.ndarray]:
         """Evalutate models until results are idempotent.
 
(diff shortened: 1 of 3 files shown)


jonmaddock:
@j-a-foster @chris-ashe can you check that you can't reproduce this problem any more? I can't seem to reproduce it, but I've imitated it to test this solution.

process/caller.py
```diff
@@ -41,7 +43,36 @@ def check_agreement(
...
+        :type previous: dict[str,float]
+        :param current: variables and values from current models evaluation
+        :type current: dict[str,float]
+        :return: non-converged values
+        :rtype: dict[str, list[float, float]
+        """
+        nonconverged_vars = {}
+        for var in previous.keys():
+            if np.allclose(
+                previous[var], current.get(var, np.nan), rtol=1.0e-6, equal_nan=True
```
timothy-nunn, line 68: I wonder whether having NaN's not equal each other will be helpful. Then, you get all the NaNs output in the terminal:

```
Variable      Previous value    Current value
----------  ----------------  ---------------
coe                      nan              nan
```

process/caller.py
```diff
@@ -158,17 +190,32 @@ def call_models_and_write_output(self, xc: np.ndarray, ifail: int) -> None:
...
                 "Model evaluations at the current optimisation parameter vector "
                 "don't produce idempotent values in the final output."
             )
+
+            print(
+                tabulate(
+                    [[k, v[0], v[1]] for k, v in nonconverged_vars.items()],
+                    headers=["Variable", "Previous value", "Current value"],
+                )
+            )
```
timothy-nunn, line 209: Should we consider adding some colour here? It can be quite hard to see these warnings and tables in all of the other PROCESS junk (as seen below)

```
...
*Cryostat roof allowance includes uppermost PF coil and outer thermal shield.
*Cryostat floor allowance includes lowermost PF coil, outer thermal shield and gravity support.
This is longer than 110 columns.
 Warning in routine OCMMNT :

*Cryostat roof allowance includes uppermost PF coil and outer thermal shield.
*Cryostat floor allowance includes lowermost PF coil, outer thermal shield and gravity support.
This is longer than 110 columns.
 Warning in routine OCMMNT :

*Cryostat roof allowance includes uppermost PF coil and outer thermal shield.
*Cryostat floor allowance includes lowermost PF coil, outer thermal shield and gravity support.
This is longer than 110 columns.
/home/ubuntu/PROCESS/process/caller.py:197: UserWarning: Model evaluations at the current optimisation parameter vector don't produce idempotent values in the final output.
  warnings.warn(
Variable      Previous value    Current value
----------  ----------------  ---------------
coe                      nan              nan
 Warning in routine OCMMNT :

*Cryostat roof allowance includes uppermost PF coil and outer thermal shield.
*Cryostat floor allowance includes lowermost PF coil, outer thermal shield and gravity support.
This is longer than 110 columns.
...
```

process/caller.py
```diff
@@ -41,7 +43,36 @@ def check_agreement(
...
         """
-        return np.allclose(previous, current, rtol=1.0e-6)
+        # Check for same shape: mfile length can change between iterations
+        if isinstance(previous, float) or previous.shape == current.shape:
+            return np.allclose(previous, current, rtol=1.0e-6, equal_nan=True)
+        else:
+            return False
+
+    @staticmethod
+    def check_mfile_agreement(
```
timothy-nunn, line 53: Can this method replace the `check_agreement` method too? We seem to have two very similar static methods used at different points in the code.

timothy-nunn (changes requested):
I agree with @j-a-foster that putting this warning and the table in the OUT.DAT (and maybe the MFILE.DAT) could be quite a good idea too.

process/caller.py
jonmaddock: Yes, I agree it's worth highlighting `nan`s, but this table is for non-idempotent values; the `nan` is idempotent in this case.

process/caller.py
jonmaddock: Yellow warning colour added.

jonmaddock:
> I've tested two input files on this branch, that fail on main, and they now pass. I can't see the warning in the OUT.DAT though?
Here's the new OUT.DAT output.
```
# ^Numerics 
   26  TF coil conduit stress upper lim      <  7.5000E+08 Pa           1.1159E+08 Pa        
 
 ****************************************** NON-IDEMPOTENT VARIABLES ******************************************
 
 Model evaluations at the current optimisation parameter vector don't produce idempotent values in the final output.
Variable      Previous value    Current value
----------  ----------------  ---------------
rmajor                     1                8
 
 *************************************** Power Reactor Costs (1990 US$) ***************************************
```
I've decided not to put anything in the MFILE output, because I don't think it's the right place for error messages.

process/caller.py
```diff
@@ -158,17 +176,36 @@ def call_models_and_write_output(self, xc: np.ndarray, ifail: int) -> None:
...
                 "Model evaluations at the current optimisation parameter vector "
                 "don't produce idempotent values in the final output."
             )
+            non_idempotent_table = tabulate(
+                [[k, v[0], v[1]] for k, v in nonconverged_vars.items()],
+                headers=["Variable", "Previous value", "Current value"],
+            )
+
+            warnings.warn("\033[93m" + non_idempotent_warning)
+            print(non_idempotent_table + "\033[0m")
```
timothy-nunn, line 195: Why is all of the output not handled by the `warn` function?

timothy-nunn (approved):
Happy with the changes. Curious why there is a `warn` and `print` statement each handling half of the out. Maybe consider changing this if its not necessary.

process/caller.py
jonmaddock: Quite right, I've combined these.

PR #3362, opened 2024-10-24
Use different GitHub API with greater file number limit

timothy-nunn:
The current regression test uses a GitHub API with a limit of 1000 files... the tracking repo now has more than that. Uses the trees API instead.

### DIFF
--- a/tests/regression/regression_test_assets.py
+++ b/tests/regression/regression_test_assets.py
@@ -21,9 +21,8 @@ class TrackedMFile:
 
 
 class RegressionTestAssetCollector:
-    remote_repository_url = (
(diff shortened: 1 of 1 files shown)


timothy-nunn:
On main the regression tests show:
```
============= 4 passed, 2 skipped, 7 warnings in 271.66s (0:04:31) =============
```

On this branch, the regression tests show:
```
================== 6 passed, 15 warnings in 307.22s (0:05:07) ==================
```

The two skipped jobs on main arise from the regression tests not seeing stellarator reference MFiles as these exist beyond the 1000 file cutoff.

PR #3369, opened 2024-10-31
Convert ife.f90 to python

timothy-nunn:
Converts the `ife_module` code into a Python model in the `process` package.


timothy-nunn:
Both regression tests fail with the following message:

```
FAILED tests/regression/test_process_input_files.py::test_input_file[IFE] - AssertionError:  MFile contains variables that are not present in the reference MFILE: {'shdzu', 'fwdzu', 'v1dzu', 'v3dzl', 'chdzl', 'v2dzu', 'bldzl', 'v1dzl', 'chdzu', 'fwdzl', 'v3dzu', 'bldzu', 'shdzl', 'v2dzl'} 
```

This is because I simplified the writing out of the vertical build which means that some more variables are written out for this IFE type

jonmaddock (approved):

PR #3377, opened 2024-11-04
Remove developer group from docs

jonmaddock:
I've decided to remove the "developer group" in the docs, as suggested in the issue. The `CITATION.cff` will be used as the full citation for the repository (with credit to all authors), whilst the "contacts" at the bottom of the readme will be the key contacts. As Graeme points out, the post-gitlab contributors are recorded by Github anyway.


ym1906 (approved):
Seems sensible to me.

PR #3392, opened 2024-11-13
Convert stellarator_configuration to Python

timothy-nunn:
Convert stellarator_configuration (except module variables) to Python


source/fortran/stellarator_variables.f90
```diff
@@ -115,5 +115,7 @@ subroutine init_stellarator_variables
     vportamax = 0.0D0
     vportpmax = 0.0D0
     vporttmax = 0.0D0
+    max_gyrotron_frequency = 1.0D9
+    te0_ecrh_achievable = 1.0D2
```
timothy-nunn, line 119: @stuartmuldrew I added these so I could run yours and Felix's old file. Are you happy with these defaults?

timothy-nunn: Previously, these values defaulting to 0 resulted in `Ignition point: Heating Power (MW)` being written to MFile as `NaN` (I presume this was ok)... however now it causes an error in the code and stops the MFile being written

jonmaddock (approved):

PR #3406, opened 2024-11-22
Convert process initialisation routines to python

timothy-nunn:
PROCESS initialisation sequence Fortran -> Python

Iteration variable initialisation sequence Fortran -> Python

`initial.f90` post-input pre-model call validation Fortran -> Python

The input validations raise a `ProcessValidationError`, a new exception introduced specifically for input validation errors in a first move away from the in-built error handling system. As a temporary measure, the non-level 3 errors use `warnings.warn` until a permanent warning solution is found.


jonmaddock (approved):
Great, this is a nice solution.

PR #3418, opened 2024-12-11
Replace maths library with Python libraries

timothy-nunn:
Removing some of the maths routines that can be replaced with simple routines from libraries like numpy and scipy.


process/pfcoil.py
```diff
@@ -964,13 +965,10 @@ def solv(self, ngrpmx, ngrp, nrws, gmat, bvec):
...
         """
-        truth = True
-        eps = 1.0e-10
         ccls = np.zeros(ngrpmx)
+        work2 = np.zeros(ngrpmx)
 
-        sigma, umat, vmat, ierr, work2 = ml.svd(
-            nrws, np.asfortranarray(gmat), truth, truth
-        )
+        umat, sigma, vmat = svd(gmat)
```
timothy-nunn, line 971: The changes to PF coil require a bit of explaining. The single value decomposition from scipy returns different results to the maths library implementation. This causes the integration tests to fail, but none of the regression tests.

In some cases, like `test_solv` the values of `ccls` changed, but their sum did not. 
In `test_efc`, the `ccls` values changed (one of the PF coils groups ended up with 0 current!?) but the `ssq` value did not, weirdly, `ssq` is calculated from the `ccls`.

Overall, I suspect this is nothing to worry about, but its worth careful consideration before approving.

tests/integration/test_pfcoil_int.py
```diff
@@ -2507,13 +2486,13 @@ def test_peakb(monkeypatch: pytest.MonkeyPatch, pfcoil: PFCoil):
...
-            -5.5984385047179153,
-            -5.5984385047179153,
-            -186.98751599968148,
+            14.742063826112622,
+            20.032681634901664,
+            0.58040662653667285,
+            0.58040662653667285,
+            0.42974674788703021,
+            0.42974674788703021,
+            174.22748790786324,
```
jonmaddock: I'm sure this is intentional, but can you explain why `curpfb` and `curfps` values have been swapped?

tests/integration/test_pfcoil_int.py
```diff
@@ -446,12 +446,13 @@ def test_efc(pfcoil: PFCoil, monkeypatch: pytest.MonkeyPatch):
...
-        579111.6216917,
-        20660782.82356247,
-    ])
+    assert ccls[0:3] == pytest.approx(
+        np.array([
+            12846165.42893886,
+            16377261.02000236,
+            579111.6216917,
+        ])
+    )
```
jonmaddock, line 455: What's the reason for removing the 4th element?

tests/integration/test_pfcoil_int.py
```diff
@@ -1642,31 +1643,9 @@ def test_solv(pfcoil: PFCoil):
...
-            [0.40824829, -0.57735027, 0.70710678],
-        ]),
-    )
-    assert_array_almost_equal(
-        sigma, np.array([5.1279005e-16, 6.0000000e00, 0.0000000e00])
-    )
-    assert_array_almost_equal(
-        work2, np.array([-2.22044605e-16, -1.73205081e00, 0.00000000e00])
-    )
+    assert_array_almost_equal(ccls, np.array([-0.069036, 0.488642, 0.080394]))
```
jonmaddock, line 1648: These are quite different, as you say.

process/pfcoil.py
jonmaddock: I see, this is tricky. There are some quite substantial (to me) differences in the test results; we just need to be happy that this is due to differences in the SVD implementations and not any mistakes in the refactoring. It's particularly strange about the sum and `ssq` values, as you say.

process/pfcoil.py
```diff
@@ -983,15 +980,14 @@ def solv(self, ngrpmx, ngrp, nrws, gmat, bvec):
 
         # Compute currents
         for i in range(ngrp):
-            ccls[i] = 0.0e0
```
jonmaddock, line 985: Was this doing anything before? Clutching at straws!

jonmaddock (changes requested):
Great, just the question of the SVD differences. I'm not sure how serious these differences are; it looks correct, but maybe we need to convince ourselves fully.

timothy-nunn:
The SVD differences can be explained as follows. The original SVD implementation does not return a correctly factorised matrix. The following code demonstrates that when attempting to recreate the original matrix (https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.svd.html) for the `test_solv`, the outputs do not create a 3x3 matrix of all 2's when running with `OLD=True`, but do when running with `OLD=False` (the new integration test values).

```python
import numpy as np

OLD = True

if OLD:
    U = np.array([
        [-0.81649658, -0.57735027, 0.0],
        [0.40824829, -0.57735027, -0.70710678],
        [0.40824829, -0.57735027, 0.70710678],
    ])
    V = np.array([
        [-0.81649658, -0.57735027, 0.0],
        [0.40824829, -0.57735027, -0.70710678],
        [0.40824829, -0.57735027, 0.70710678],
    ])
    s = np.array([5.1279005e-16, 6.0000000e00, 0.0000000e00])
else:
    U = np.array([
        [-0.57735027, -0.57735027, -0.57735027],
        [-0.57735027, -0.21132487, 0.78867513],
        [-0.57735027, 0.78867513, -0.21132487],
    ])
    V = np.array([
        [-0.57735027, -0.57735027, -0.57735027],
        [0.0, -0.70710678, 0.70710678],
        [0.81649658, -0.40824829, -0.40824829],
    ])
    s = np.array([6.0, 0.0, 0.0])


# reconstruct
sigma = np.zeros((3, 3))
for i in range(3):
    sigma[i, i] = s[i]

a1 = np.dot(U, np.dot(sigma, V))

print(f"Checking {'old' if OLD else 'new'} SVD results")
print("Original matrix was 3x3 all 2's")
print(a1)
```

tests/integration/test_pfcoil_int.py
timothy-nunn: Im not sure this is happening anymore

tests/integration/test_pfcoil_int.py
timothy-nunn: It became 0, I will add it back since its probably correct to do so

process/pfcoil.py
timothy-nunn: Relic of Fortran

jonmaddock (approved):

PR #3426, opened 2024-12-13
Convert main module to python

timothy-nunn:
Convert `main_module.f90` to Python. 

I have chosen not to output the `commsg` (commit message) anymore to the MFile, hence the regression failures. This is because I do not believe the information is useful, its often truncated, and it should be possible to find the commit message by the commit hash included in the Git tag.


timothy-nunn:
Draft until #3406

jonmaddock (approved):

PR #3478, opened 2025-01-15
Fix and enable all RUF rules

timothy-nunn:


process/pfcoil.py
```diff
@@ -2223,15 +2223,13 @@ def outpf(self):
                 if (
                     (pfv.isumatoh == 6 or pfv.isumatoh == 8 or pfv.isumatoh == 9)
                     and tfv.str_cs_con_res > 0.7e-2
-                    or tfv.str_cs_con_res < -0.7e-2
-                ):
+                ) or tfv.str_cs_con_res < -0.7e-2:
```
je-cook, line 2226: is this the right way round?  I would have naiively thought it would be the (or first three) and (4 or 5). Same for the second

je-cook (approved):
Just the one question otherwise good to go

process/pfcoil.py
timothy-nunn: I have checked and changed it to just compare with the `abs`

PR #3486, opened 2025-01-17
Enable "G" rule in ruff

jonmaddock:


timothy-nunn:
Didn't quite appreciate that all the G failures were 004 which I have since ignored :D

PR #3495, opened 2025-01-22
enabling ARG rule in ruff

clmould:
## Checklist


timothy-nunn (approved):
I think this solution is appropriate until a restructure of the models. Some of this stuff looks to be the result of interfaces that require certain arguments even if not always used. 

Please take a look at the unit tests because they probably use named arguments which have now changed

timothy-nunn (changes requested):
Clicked wrong button before 🤦. Please address at test failures

process/io/mfile.py
```diff
@@ -131,7 +131,7 @@ def __init__(self, item):
         self.set_scan = self.get_error
         self.get_number_of_scans = self.get_error
 
-    def get_error(self, *args, **kwargs):
+    def get_error(self, *args, **kwargs):  # noqa: ARG002
```
timothy-nunn, line 134: No need to exclude this, just remove the arguments from the associated calls

timothy-nunn: I am wrong here. `MFileErrorClass` needs to act like the `MFile` class but does not explicitly inherit an interface to make this clear. I think it is correct to ignore `ARG002` here

timothy-nunn (approved):

PR #3537, opened 2025-02-11
Replaced stellarator only variable with tokamak used variable

### LINKED ISSUE #3109: coolant channels within the tokamak and stellarator models
Within Process there are variables for the radius of the cooling channels that run through components that need cooling.

In particular  rcool for the centrepost

For the first wall and blankets it is afw used for inboard and outboard

There is also afwi and afwo variables that have been introduced for the inboard and outboard use of the cooling channels within the first wall and the blankets. But they only seem to be used for the stellarator model.

Surely it would be better if they are properly used in tokamak and stellarator models? 

possibly it would be better to have afwi and afwo for the inboard and outboard cooling radiuses of the first wall cooling pipes

and maybe a new variable for the inboard and outboard cooling pipes for the blankets? i.e  blk_rci and blk_rco

It looks as if it would not be that hard to do.

kj5248:
Closes #3109 

## Description

In stellarator there were variables afwi and afwo for the coolant inner channel radius for the FW inboard and outboard respectively. These variables seemed to have the same value in all the stellarator files and is comparable to the tokamak variable radius_fw_channel which is the inner channel radius for the FW (inboard and outboard).

For this reason, and to also allow for the potential of using the dcll blanket model with the stellarator, these variables were replaced with the tokamak variable radius_fw_channel.

Also had to make a change to the stellarator regression file as the default value for afwi and afwo was 0.008 rather than the default of radius_fw_channel which is 0.006.

## Checklist


kj5248:
Was easier to make the changes in a fresh branch than resolving the merge conflicts from the various rebasing, so this will be closed and is resolved in #3580

PR #3552, opened 2025-02-19
Convert inputf.90 to python

timothy-nunn:
Converts PROCESS input file parsing, validation, and actioning (mostly setting variables) to Python

## Limitations
- Cannot parse string arrays (might be possible but I do not think its done in the code)

## Notes
- Did not include `factor` as an input anymore, the variable should be removed.


timothy-nunn:
Further evidence I have observed that our old parsing was different (wrong):

Using the input.f90 `string_to_real` method reduced the number of failures in the once through regression tests. The new method casts the string in Python and so we should probably trust the results of that?

process/input.py
```diff
@@ -129,27 +1664,50 @@ def parse_input_file():
...
-            # TODO: re-enable error instead
-            # raise ProcessValidationError(error_msg)
+            raise ProcessValidationError(error_msg)
 
         # validate the variable and also clean it (cast to correct type)
-
-        clean_variable_value = validate_variable(
-            variable_name, variable_value.strip(), array_index, variable_config, line_no
-        )
+        if "," in variable_value:
```
jonmaddock, line 1670: Comment to make this clearer please.

process/input.py
```diff
@@ -163,17 +1721,28 @@ def parse_input_file():
             variable_config.additional_actions(
                 variable_name,
                 clean_variable_value,
-                None if array_index is None else int(array_index),
+                array_index_clean,
                 variable_config,
             )
 
+        # again its the input name not the target name
```
jonmaddock, line 1729: Can this comment be clarified?

process/input.py
```diff
@@ -185,7 +1754,7 @@ def validate_variable(
     :returns: the input value with the correct type.
     """
     # check its an array if and only if the config says so
-    if array_index is None and config.array:
+    if array_index is None and config.array and check_array:
```
jonmaddock, line 1758: When would you not want to check an array?

scripts/create_dicts.py
```diff
@@ -764,6 +764,7 @@ def dict_input_bounds():
     for var_name, config in INPUT_VARIABLES.items():
         lb, ub = None, None
         if config.range is not None:
+            print(var_name)
```
jonmaddock, line 767: Debug?

process/init.py
```diff
@@ -441,24 +441,25 @@ def check_process(inputs):
...
-    if fortran.fwbs_variables.iblanket_thickness == 1:
-        fortran.build_variables.dr_blkt_inboard = 0.53
-        fortran.build_variables.dr_blkt_outboard = 0.91
-    elif fortran.fwbs_variables.iblanket_thickness == 2:
-        fortran.build_variables.dr_blkt_inboard = 0.64
-        fortran.build_variables.dr_blkt_outboard = 1.11
-    elif fortran.fwbs_variables.iblanket_thickness == 3:
-        fortran.build_variables.dr_blkt_inboard = 0.75
-        fortran.build_variables.dr_blkt_outboard = 1.30
+    if "iblanket_thickness" in inputs:
```
jonmaddock, line 444: Why does this have to be treated differently?

tests/unit/test_input.py
```diff
@@ -71,6 +71,32 @@ def test_parse_real(epsvmc, expected, tmp_path):
...
+    [
+        [0.546816593988753],
+        [0.13134204235647895],
+        [0.75],
+        [0.7],
+        [0.3],
+        [0.1293140904093427],
+    ],
+)
+def test_parse_real_didnt_work_in_f90(value, tmp_path):
```
jonmaddock, line 85: Ok, but should this be a test now?

jonmaddock (changes requested):
This is great Tim, SO much better. Just some clarifying comments required please. Did you check out cerberus for this?

tests/unit/test_input.py
timothy-nunn: Renamed the test, I think its still valuable

process/init.py
timothy-nunn: This is to ensure the code's behaviour does not change as a result of this conversion.

https://github.com/ukaea/PROCESS/blob/10691237047d7d0ffbe97d55a42c8875ef898505/source/fortran/input.f90#L2233-L2245

When I did not perform this check, it changed the results of the tests!

process/input.py
timothy-nunn: This is actually no longer necessary and is left over from how I originally implemented full arrays, removed

jonmaddock (approved):
Great, I'm happy with this. Did you look into cerberus?

timothy-nunn:
> Great, I'm happy with this. Did you look into cerberus?

@jonmaddock I didn't end up using cerberus because the validation we implement here is very minimal and existing at the time of investigation. In future, it will be worth looking at this as we convert our data structure and IO formats.

PR #3553, opened 2025-02-20
Add single model evaluation example notebook

jonmaddock:
Add an example notebook demonstrating single-model evaluation in Process, without using the CLI and input/output files.


examples/single_model_evaluation.ipynb
```diff
@@ -0,0 +1,409 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 15,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "pdivt = 42.364607206184246\n"
```
timothy-nunn, line 41: This should be `0`? Does the notebook need to be re-run from scratch and re-pushed?

examples/single_model_evaluation.ipynb
```diff
@@ -0,0 +1,409 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def run_impurities(w_imp_fracs):\n",
+    "    \"\"\"Calculate responses to W impurities.\"\"\"\n",
+    "    n = w_imp_fracs.shape[0]\n",
```
timothy-nunn, line 265: I think commenting that `n` here is the number of times we change the W frac and observe the responses helps make this clearer quicker.

examples/single_model_evaluation.ipynb
```diff
@@ -0,0 +1,409 @@
...
+    "    p_plasma_rad_mw = np.empty(n)\n",
+    "    pdivt = np.empty(n)\n",
+    "    p_l_h_threshold_mw = np.empty(n)\n",
+    "    con15 = np.empty(n)\n",
```
timothy-nunn, line 269: Is it necessary to do this, or can we just make these as normal arrays and append the values?

Then comment on the loop that we are iterating over the W frac test points and storing key physical responses at each test point.

examples/single_model_evaluation.ipynb
```diff
@@ -0,0 +1,409 @@
...
+    "\n",
+    "fig, ax = plt.subplots()\n",
+    "ax.scatter(w_imp_fracs, p_plasma_rad_mw, label=\"p_plasma_rad_mw\")\n",
+    "ax.scatter(w_imp_fracs, pdivt, label=\"pdivt\")\n",
+    "ax.plot(w_imp_fracs, p_l_h_threshold_mw, \"r\", label=\"p_l_h_threshold_mw\")\n",
+    "ax.ticklabel_format(style=\"sci\", axis=\"x\", scilimits=(0, 0))\n",
+    "ax.set_title(\"W impurity fraction against radiated and divertor power\")\n",
+    "ax.set_xlabel(\"W impurity fraction\")\n",
+    "ax.set_ylabel(\"Power (MW)\")\n",
+    "ax.legend()"
```
timothy-nunn, line 330: This didn't work for me without adding `%matplotlib inline` to the top of the cell

examples/single_model_evaluation.ipynb
```diff
@@ -0,0 +1,409 @@
...
+   "source": [
+    "fig, ax = plt.subplots()\n",
+    "ax.scatter(w_imp_fracs, con15, label=\"con15\")\n",
+    "ax.ticklabel_format(style=\"sci\", axis=\"x\", scilimits=(0, 0))\n",
+    "ax.set_title(\"W impurity fraction against L-H threshold constraint (15)\")\n",
+    "ax.set_xlabel(\"W impurity fraction\")\n",
+    "ax.set_ylabel(\"L-H threshold constraint value\")\n",
+    "ax.hlines(0.0, xmin=ax.get_xlim()[0], xmax=ax.get_xlim()[1], colors=\"r\")\n",
+    "ax.annotate(\"Satisfied\", (0.0, 0.1))\n",
+    "ax.annotate(\"Violated\", (0.0, -0.15))"
```
timothy-nunn, line 375: Same as above, figures did not display until `%matplotlib inline`

timothy-nunn (changes requested):
Couple of issues probably caused by different versions of Jupyter/matplotlib.

Is it worth noting somewhere that this notebook will only become easier as the Pythonisation continues? I'm in two minds, and I'll leave that up to you.

.gitignore
```diff
@@ -62,3 +62,4 @@ python_modules.txt
 quench_data.DAT
 env-fork
 *SIG_TF.json
+!examples/data/*
```
timothy-nunn, line 65:
```suggestion
!examples/data/*IN.DAT
```

Otherwise the MFILE and OUT file become untracked when they should not be

examples/single_model_evaluation.ipynb
jonmaddock: Good idea, done.

examples/single_model_evaluation.ipynb
jonmaddock: Yeah, maybe a version thing as you say. I've add it.

examples/single_model_evaluation.ipynb
jonmaddock: Same as above.

.gitignore
jonmaddock: True, but we have other output files in `examples/data` that we want to be tracked, e.g. the ones we use in `plot_solutions.py`. I could ignore the ones associated with this notebook specifically if you like?

examples/single_model_evaluation.ipynb
jonmaddock: I did this because using `append()` is inefficient; you're making an array copy with every operation/iteration. If you know the length beforehand, I think it's more efficient to fill-in a pre-defined empty array. Perhaps I'm missing your point here though? Not like efficiency is big deal here, but I think it's better practice.

Added a comment to clarify, good idea.

examples/single_model_evaluation.ipynb
jonmaddock: Done.

.gitignore
timothy-nunn: I believe you could also use `examples/.gitignore` to make some more specific rules away from the main .gitignore. Excluding all MFiles except the 6 meant to be there. Its not the most elegant situation whichever way we do it.

timothy-nunn (changes requested):
Thanks for actioning my comments. 

I have replied to the `.gitignore` thread. There are also some obsolete variables that need to be fixed if you could rebase onto main and correct these.

.gitignore
jonmaddock: Ok, I've added some specific negations to the `examples/.gitignore`.

examples/.gitignore
```diff
@@ -1,3 +1,6 @@
 plot_proc_1.png
 plot_proc_2.png
 plot_proc_3.png
+!large_tokamak_?_MFILE.DAT
+!csv_output_large_tokamak_MFILE.DAT
+!large_tokamak_once_through_MFILE.DAT
```
timothy-nunn, line 6: I don't think we want to exclude the once-through MFile, right?

jonmaddock: Urgh, whoops! I've it changed to to negate `large_tokamak_once_through_IN.DAT` instead.

PR #3570, opened 2025-03-06
Convert iteration_variables.f90 to Python

timothy-nunn:
Convert `iteration_variables.f90` to Python.

Similar to #3552, I have created a config class and created an entry (in a dict this time) for each existing iteration variable.

### DIFF
--- a/process/iteration_variables.py
+++ b/process/iteration_variables.py
@@ -1,0 +1,223 @@
+from dataclasses import dataclass
+from typing import Any
+
+import process.fortran as fortran
+
+
+@dataclass
+class IterationVariable:
+    name: str
+    """The name of the variable"""
+    module: Any
+    """The Fortran module that this variable should be set on."""
+    lower_bound: float
+    """The default lower bound of the iteration variable"""
+    upper_bound: float
+    """The default upper bound of the iteration variable"""
+
+
+ITERATION_VARIABLES = {
+    1: IterationVariable("aspect", fortran.constants, 1.100e0, 10.00e0),
+    2: IterationVariable("bt", fortran.build_variables, 0.010e0, 30.00e0),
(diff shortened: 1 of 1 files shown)


timothy-nunn:
This section of code could have been simplified
https://github.com/ukaea/PROCESS/blob/159ac627b8054dcf11dd02e64345a3bca1ba5698/source/fortran/iteration_variables.f90#L4202-L4210

$s=1/x$
and then later on
$x=x\times s$

so it follows that when $x$ is set at the end of the snippet, it is to $1$. While this is correct, it causes a difference in the variable value that causes differences in once-throughs. My guess is that this happens because of some floating point error that accumulates. As a result, some weird expressions exist within the code that look like they should have been simplified.

timothy-nunn:
Unsure why the ST regression test is failing. I get no differences:
```
$ diff old.st_regression.MFILE.DAT new.st_regression.MFILE.DAT 
7c7
<  Time_of_run_____________________________________________________________ (time)________________________     "11:45"
---
>  Time_of_run_____________________________________________________________ (time)________________________     "11:38"
10,11c10,11
<  PROCESS_git_tag_________________________________________________________ (tagno)_______________________     "v3.1.0-314-g159ac627"
<  PROCESS_git_branch______________________________________________________ (branch_name)_________________     "main"
---
>  PROCESS_git_tag_________________________________________________________ (tagno)_______________________     "v3.1.0-319-gb585968d"
>  PROCESS_git_branch______________________________________________________ (branch_name)_________________     "3432-convert-iteration_variablesf90-to-python"
```

jonmaddock (changes requested):
Due to my inaction on this, there are now some conflicts. Could you resolve these before I review?

jonmaddock (approved):
Great, nice to get the dicts working nicely too.

PR #3579, opened 2025-03-11
Continuous kludge to keep pdivt > 0

jonmaddock:
Ensure `pdivt` is kept positive continuously.
### `pdivt` with continuous kludge:
kludged_pdivt
### And again, but with finer sampling resolution to show continuous nature:
kludged_pdivt_zoom

### DIFF
--- a/process/physics.py
+++ b/process/physics.py
@@ -2240,9 +2240,12 @@ class Physics:
             - physics_variables.p_plasma_rad_mw
         )
 
-        # The following line is unphysical, but prevents -ve sqrt argument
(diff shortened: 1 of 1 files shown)


jonmaddock:
> I am bit concerned that you are redefining `pdivt`. If this code is ever run again then this change could be made repeatedly. Is there any chance of this happening?

I'm not quite sure what you mean; if you mean non-idempotence, then this isn't an issue: each time the model is called (with the same values) `pdivt` is calculated from other values (in the lines immediately above this change) and this kludge is applied. The same value of `pdivt` results each time.

mkovari (approved):

PR #3582, opened 2025-03-12
Added new functionality to plot_scans.py allowing axis tick labels to be percentages rather than absolutes

### LINKED ISSUE #1875: Add option to plot percentage to plot_scans.py
In GitLab by @j-a-foster on Jun 12, 2023, 16:31

It would be useful to have the option to change a plot axis to percentages instead of absolute values. I've already got `plot_scans.py` to do it for a specific case and it would be easy to implement more generally:

```python
ax.set_yticks(list_of_ticks)
ax.yaxis.set_major_formatter(mtick.PercentFormatter(max_value))
```
where `list_of_ticks` can be calculated based on what `max_value` is. Could possibly provide option to not have max be 100% but not sure how useful this would be.

kj5248:
closes #1875 

## Description

Added two new arguments to plot_scans.py: "--x_axis_percent" ("-x%") and "--y_axis_percent" ("-y%").

These will change the axis tick labels from absolute values to percentage values (based upon the maximum value from MFILE).

This was checked for the different plot options available in plot_scans:

[2Dcontour.pdf](https://github.com/user-attachments/files/19206422/2Dcontour.pdf)
[2dplot.pdf](https://github.com/user-attachments/files/19206423/2dplot.pdf)
[plot.pdf](https://github.com/user-attachments/files/19206424/plot.pdf)
[stackplot.pdf](https://github.com/user-attachments/files/19206425/stackplot.pdf)
[twinaxisplot.pdf](https://github.com/user-attachments/files/19206426/twinaxisplot.pdf)

Additionally, the tick label size was not functional for all plots so switched out the command used so that it now works, also added a new argument "--axis_ticklabel_size" ("-ats") that allows for it to be varied from the default, necessary for some cases when using percentages for the x axis.

Also added arguments: "--x_axis_range", "--x_axis_max", "--y_axis_range", "--x_axis_max" to allow for more options for graphs, these work with the percentage axes turned on or off. For the percentages cases the max and range values are as a percent rather than an absolute.

## Checklist

### DIFF
--- a/process/io/plot_scans.py
+++ b/process/io/plot_scans.py
@@ -570,7 +570,7 @@ def main(args=None):
                         y_divisions = (y_axis_range[1] - y_axis_range[0]) / 10
                     if y_axis_percentage:
                         if y_max is None:
-                            y_max = max(abs(output_arrays[input_file][output_name]))
+                            y_max = max(np.abs(output_arrays[input_file][output_name]))
                         yticks = mtick.PercentFormatter(y_max)
                         if y_axis_range != []:
                             y_divisions = 5 * math.ceil(y_divisions / 5) * y_max / 100
@@ -586,7 +586,7 @@ def main(args=None):
                         x_divisions = (x_axis_range[1] - x_axis_range[0]) / 10
                     if x_axis_percentage:
                         if x_max is None:
-                            x_max = max(abs(scan_var_array[input_file]))
+                            x_max = max(np.abs(scan_var_array[input_file]))
                         xticks = mtick.PercentFormatter(x_max)
                         ax.xaxis.set_major_formatter(xticks)
                         if x_axis_range != []:
@@ -615,7 +615,9 @@ def main(args=None):
                             y_divisions = (y_axis_range[1] - y_axis_range[0]) / 10
                         if y_axis_percentage:
                             if y_max is None:
-                                y_max = max(abs(output_arrays[input_file][output_name]))
+                                y_max = max(
+                                    np.abs(output_arrays[input_file][output_name])
+                                )
                             yticks = mtick.PercentFormatter(y_max)
                             if y_axis_range != []:
                                 y_divisions = (
@@ -639,7 +641,7 @@ def main(args=None):
                             x_divisions = (x_axis_range[1] - x_axis_range[0]) / 10
                         if x_axis_percentage:
                             if x_max is None:
-                                x_max = max(abs(scan_var_array[input_file]))
+                                x_max = max(np.abs(scan_var_array[input_file]))
                             xticks = mtick.PercentFormatter(x_max)
                             print("xticks", xticks)
                             if x_axis_range != []:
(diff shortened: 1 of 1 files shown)


j-a-foster (changes requested):
Currently, the `--y_axis_max` argument affects both y axes if you plot two. There should probably be a `--y_axis_max` and `--y_axis_2_max` or something. There are `-yv` and `-yv2` arguments so the max arguments should reflect these.

process/io/plot_scans.py
```diff
@@ -444,6 +549,12 @@ def main(args=None):
         # Plot section
```
timothy-nunn, line 549: Move this section (until the `for` loop) to the top of the `for` loop. This will stop it from only plotting the first figure and leaving the rest blank

timothy-nunn (changes requested):
Mainly minor quality changes. I found the reason why only the first plot was being plot and have added it as a comment below.

process/io/plot_scans.py
```diff
@@ -486,6 +596,44 @@ def main(args=None):
                         color="blue" if len(input_files) == 1 else None,
                         label=labl,
                     )
+                    if y_axis_range != []:
+                        y_divisions = (y_axis_range[1] - y_axis_range[0]) / 10
+                    if y_axis_percentage:
+                        if y_max is None:
+                            y_max = max(np.abs(output_arrays[input_file][output_name]))
```
timothy-nunn, line 603: Overwriting `y_max` here causes any plot produced after the first to have the wrong percentage. Because `y_max` is only `None` for the first plot and keeps the same value afterwards. This caused some of the later plots to have thousands of %. 

This seems to happens four times with `y_max` and at least once with `y_max2` can you ensure we do not overwrite variables input from the CLI.

timothy-nunn (changes requested):

process/io/plot_scans.py
timothy-nunn: You can see this happening in `examples/scan.ipynb`

kj5248: Thanks for pointing this out, it should be resolved now. Not sure it is the best way of doing this? But it works and shouldn't cause any issues.

kj5248: The max values for the twinned axis is not working. I will fix it when I am back from AL later this week.

timothy-nunn (changes requested):
Just two remaining print statements and its good

process/io/plot_scans.py
kj5248: Now fixed along with others...Have found that it will only work if you give a single -yv2 variable not sure if this is by design or an error. This seems to also be the case on main so might write up a separate issue for that. If you use either -yr or -yr2 without the other set the axis do weird things: 

[scan_bmaxtf_vs_coe_vs_rminor.pdf](https://github.com/user-attachments/files/19583581/scan_bmaxtf_vs_coe_vs_rminor.pdf)

It seems to take what should be used for the ax1 and using it for ax2 despite it being set to do it for the first not the second?

PR #3586, opened 2025-03-12
Convert model initialisation routines to python

timothy-nunn:
Converts module initialisation (setting of initial defaults) to Python.

In some cases, there was not an obvious places for these functions to live and so I just added to the file that seemed to use the variables the most.

See comment below for the annoying reason there are some slight test changes despite this being a line-for-line port of the code from Fortran to Python.


timothy-nunn:
Errors in the TF coil are being caused by setting 
```python
tfv.poisson_cond_axial = 0.3
tfv.poisson_cond_trans = 0.3
```

WHY!?

jonmaddock (approved):
Great stuff, just need to resolve conflicts then go for it!

timothy-nunn:
A similar issue to above is occurring at:
https://github.com/ukaea/PROCESS/blob/f55aca535941c6143fe9d3be2dc4c48b46b99cb3/source/fortran/pfcoil_variables.f90#L521

When I calculate `70/22` in Fortran and assign it to the `ld_ratio_cst` the large Tokamak once through test stops failing

PR #3595, opened 2025-03-19
Fix bugs in pythonised input parser

timothy-nunn:
Fixes two bugs in the input parser that impacts IFE. 
1. Capital letters are now ignored in variable names
2. 2D arrays can now be parsed

For the 2D arrays, there is a bit of messing around with the ordering because the input specifies the index of a flattened column-major array whereas numpy tries to treat it as a row-major array.


jonmaddock (approved):

timothy-nunn:
Im happy with the IFE regression failure: the new variables have not been updated because the IFE data generation has been failing on main.

PR #3596, opened 2025-03-20
Add colour to exceptions from test suite

timothy-nunn:
I decided not to use a package to do this because I don't feel the need to introduce another dependency just for 5 lines of colour. The ANSI codes should work with the CI and all terminals we use.


timothy-nunn:
image

jonmaddock (approved):

PR #3616, opened 2025-04-02
Remove level 3 errors raised in python

timothy-nunn:
Converts all Level 3 errors in Python to exceptions. Level 3 errors caused PROCESS to crash with a STOP 1, now its done the Pythonic way where possible


jonmaddock (approved):
Looks great!

PR #3619, opened 2025-04-03
Skip failing inequalities large tokamak regression test

jonmaddock:


timothy-nunn (approved):
This is a temporary measure while we investigate several solutions to this problem.

PR #3627, opened 2025-04-09
Enable net electric constraint in ST

timothy-nunn:


timothy-nunn:
@chris-ashe Ive also enabled iteration variable 26 because it looks like it should have been given we use equation 9

Issue #3574, opened 2025-03-07
Value of `dr_tf_inboard` inconsistent in MFILE


jonmaddock:
When selected as an optimisation parameter, the solution value of `dr_tf_inboard` in the MFILE is inconsistent with its `itvarXXX` value.
```
 dr_tf_inboard____________________________________________________________ (itvar007)_____________________ 1.19999999999999996e+00
 dr_tf_inboard_(final_value/initial_value)________________________________ (xcm007)_______________________ 1.00000000000000000e+00
 dr_tf_inboard_(range_normalised)_________________________________________ (nitvar007)____________________ 1.16279069767441845e-01
...
 TF_coil_inboard_leg_radial_thickness_(m)_________________________________ (dr_tf_inboard)________________ 7.61474358840885657e-01
```
The optimisation parameter solution value (`itvar007`) appears to still be at its initial value (see `xcm007` == 1.0). This needs to be reconciled.

clmould:
https://github.com/ukaea/PROCESS/blob/aac5f4d7a72f004ea0fb65bbd5bf86bd47bb9963/process/build.py#L1684-L1704
The large tokamak file has both ixc = 140 and ixc = 13. Having 140 enabled causes the value of dr_tf_inboard (which is ixc=13) to be overwritten.

mkovari:
The [docs](https://ukaea.github.io/PROCESS/eng-models/tf-coil/#tf-coil-inboard-radial-size) say 

> Direct parametrization: the TF radial inboard thickness width is set as an input variable : dr_tf_inboard (iteration variable 57).

Iteration variable 57 is actually 'dr_tf_nose_case'.  This error in the docs may not be relevant to the issue, but it seems worth fixing.

mkovari:
@ajpearcey Alex, You might want to think about how you would like to set up the TF coil geometry in the DEMO and large tokamak input files.

clmould:
@ajpearcey one idea we had is to remove ixc = 13 from the LT in.dat files since this variable is being overwritten, do you think this would make sense?

jonmaddock:
Interestingly, @timothy-nunn also found this in #3654 . It is my view that the model should never be aware of what optimisation parameters are active; this means the model is dependent on the type of problem being solved.

clmould:
This issue will be solved by #3654

PR #3646, opened 2025-04-29
Solve equality constraints when evaluating

jonmaddock:
Implements `scipy`'s `fsolve` to solve equality constraints when evaluating only (not optimising).


tests/regression/input_files/spherical_tokamak_once_through.IN.DAT
```diff
@@ -28,76 +32,12 @@ icc = 16 * Net electric lower limit
...
-ixc = 110 * ftaulimit
-boundl(110) = 0.01
-boundu(110) = 1.0
-ixc = 116 * fradwall
-boundl(116) = 0.01
-boundu(116) = 1.0
-ixc = 135 * fimp(13)
-ixc = 140 * dr_tf_wp
-boundl(140) = 0.4
-boundu(140) = 0.8
```
timothy-nunn, line 91: Removal of this constraint causes constraint-aware code to not be executed hence the divide by 0 :/  https://github.com/ukaea/PROCESS/blob/b5572ebaa54827dca00b2170b7a31886ad326e28/process/build.py#L1689

tests/regression/test_process_input_files.py
```diff
@@ -278,6 +278,10 @@ def test_input_file(
         pytest.skip(
             "The inequalities large tokamak input file currently doesn't solve."
         )
+    elif input_file.name == "spherical_tokamak_once_through.IN.DAT":
+        pytest.skip(
+            "The spherical tokamak evaluation currently encounters a divide by zero."
+        )
```
timothy-nunn, line 284: Reminder to check this is unskipped once fixed

timothy-nunn (changes requested):
I would be much happier if the ST once through was able run. Hopefully that should be possible by just changing which iteration variables we choose to run for the consistency solves.

mkovari:
There are some substantial differences - one of the PF coil currents has changed sign and the superconductor mass in the TF coil has gone down by a factor of four, for example.  It's not obvious why this has happened.

I am struggling to find and understand the relevant files.  This looks like the input file: `tests/regression/input_files/large_tokamak_once_through.IN.DAT`,
but it seems to have a full set of iteration variables rather than the three that are required.

process/scan.py
```diff
@@ -274,64 +291,82 @@ def post_optimise(self, ifail: int):
...
+            process_output.write(
+                constants.nout,
+                f"{msg}\n",
+            )
         else:
-            string1 = "PROCESS has tried to optimise"
+            if ifail == 1:
+                string1 = "PROCESS has successfully optimised"
+            else:
+                string1 = "PROCESS has tried to optimise"
```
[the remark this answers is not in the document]
jonmaddock: Done.

process/scan.py
```diff
@@ -455,13 +495,16 @@ def post_optimise(self, ifail: int):
...
+        # Write inequality constraints
         if numerics.nineqns > 0:
             inequality_constraint_table = []
-            process_output.osubhd(
-                constants.nout,
-                "The following inequality constraint residues should be "
-                "greater than or approximately equal to zero :",
-            )
+            # Inequalities not necessarily satisfied when evaluating
+            if self.solver != "fsolve":
```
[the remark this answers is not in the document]
jonmaddock: The inequalities are still written out, just this message about the residuals being $\ge 0$ isn't written, as it's not necessarily true; the inequalities can be violated, as only the equalities are being solved for.

tests/regression/input_files/spherical_tokamak_once_through.IN.DAT
jonmaddock: Well found! I've created #3654 to deal with this separately.

tests/regression/test_process_input_files.py
jonmaddock: This is now reminded in #3654 .

jonmaddock:
> There are some substantial differences - one of the PF coil currents has changed sign and the superconductor mass in the TF coil has gone down by a factor of four, for example. It's not obvious why this has happened.
> 
> I am struggling to find and understand the relevant files. This looks like the input file: `tests/regression/input_files/large_tokamak_once_through.IN.DAT`, but it seems to have a full set of iteration variables rather than the three that are required.

Regarding the `tests/regression/input_files/large_tokamak_once_through.IN.DAT` input file, which is the main example I've used to test this idea, this now only contains only 2 iteration variables, `te` and `dene`. This is to solve just the beta and plasma power balance equality constraints.

Regarding the other dramatic swings in output variables, I agree it's not clear why solving for the two iteration variables in the LT "once_through" case:
```
(parameter = initial value -> solution)
dene     = 7.796e+19 -> 7.747e+19
te       = 12.221 -> 12.305
```
has caused the other dramatic changes you note. However, this process is no different to current Process optimisation solves, only that now only the equality constraints are being solved for, and this behaviour is more explicit. Perhaps it's worth investigating how those models (e.g. PF and TF coils) behave for small variations in electron density and temperature for signs of model "bad behaviour" (e.g. discontinuities etc.). But I believe this is a different issue and not the fault of these changes, other than that this now made explicitly clear.

jonmaddock:
@mkovari To aid comparison of the OUT.DAT before and after this change for the large tokamak "once-through" (now "eval"):

### Before:
83410868bd7eb1855dbbc61abc09085c9e69ce31
Allow bootstrap current fraction to be set via switch ([#3647](https://github.com/ukaea/PROCESS/issues/3647))
[large_tokamak_once_through.OUT.DAT.txt](https://github.com/user-attachments/files/20062772/large_tokamak_once_through.OUT.DAT.txt)

### After:
304fd0eac184e52d681f76a50283e104504c9b60
Rename once-through to evaluation
[large_tokamak_eval.OUT.DAT.txt](https://github.com/user-attachments/files/20062785/large_tokamak_eval.OUT.DAT.txt)

documentation/proc-pages/solver/equation-solver.md
```diff
@@ -1,18 +1,14 @@
...
+Sometimes one wants to find an optimal machine that is both consistent with the physics and engineering constraints but also minimises or maximises a certain *figure of merit*. This requires running `PROCESS` in optimisation mode. For these applications `PROCESS` uses the routine `VMCON`[^1] based on a variable metric method for constrained optimisation by Powell[^2]. If finds a stationary point of an objective function/figure of merit consistent with a set of equality and inequality constraints. It is designed to solve the necessary but not sufficient conditions of a constrained optimum, hence, the solution does not have to be a global minimum.
```
ajpearcey, line 11: "If finds", should be "It finds"

documentation/proc-pages/solver/solver-guide.md
```diff
@@ -124,18 +124,44 @@ It should be remembered that the value of the scan variable is set in the array
...
 
-Non-optimisation mode is sometimes used to perform benchmark comparison, whereby the machine size, output power etc. are known and one only wishes to find the calculated stresses, beta values and fusion powers, for example.
+Evaluation mode is used to evaluate models for a given set of input parameters (a "point") whilst ensuring that the models are self-consistent. It can also be used to perform benchmark comparison, whereby the machine size, output power etc. are known and one only wishes to find the calculated stresses, beta values and fusion powers, for example.
 
-Running `PROCESS` in non-optimisation mode requires few changes to be made to the input file from the optimisation case. The main differences between optimisation mode and non-optimisation mode are:
+Running `PROCESS` in evaluation mode requires few changes to be made to the input file from the optimisation case. The main differences between optimisation mode and evaluation mode are:
 
-1. Non-optimisation mode does NOT apply lower or upper bounds to the iteration variables. It follows that limit equations are no enforced.
+1. Evaluation mode does not apply lower or upper bounds to the iteration variables.
```
ajpearcey, line 133: What about `te` or `dene` or which ever parameters are used to solve equality constraints? Does `fsolve` use the upper and lower bounds?

ajpearcey (approved):

jonmaddock:
> I think the differences arise because the old input file has `ixc = 140 * dr_tf_wp` (even though there is no optimisation) and the new file doesn't. Having 140 enabled causes the value of dr_tf_inboard (which is ixc=13) to be overwritten.
> 
> ```
> Total inboard leg radial thickness (m)          (dr_tf_inboard)                 9.26725862473977369e-01
> dr_tf_inboard    = 1.2 * inboard TF coil thickness; (centrepost for ST) (m)
> ```
> 
> ```
> Total inboard leg radial thickness (m)          (dr_tf_inboard)                 1.19999999999999996e+00
> dr_tf_inboard    = 1.2 * inboard TF coil thickness; (centrepost for ST) (m)
> ```
> 
> I knew there was something funny going on! See #3574. Please note @clmould

Thanks for this Michael; this was also found in #3654 ! So, an optimisation-parameter-aware model is the cause of the large discrepancy.

documentation/proc-pages/solver/equation-solver.md
jonmaddock: Corrected.

documentation/proc-pages/solver/solver-guide.md
jonmaddock: No, `fsolve` doesn't use any bounds on the values.

mkovari:
I think it would be good to fix the large tokamak input file at this stage.  This won't necessarily solve the problem here (or implement your view that the model should never be aware of what optimisation parameters are active), but it would mean we have a regression test that makes some sense.

jonmaddock:
> I think it would be good to fix the large tokamak input file at this stage. This won't necessarily solve the problem here (or implement your view that the model should never be aware of what optimisation parameters are active), but it would mean we have a regression test that makes some sense.
> 
> Would you prefer it if I approve this pull request first?

I think I would prefer to treat the large tokamak problems separately to this PR, as they were pre-existing:
- #3574 Mutually exclusive optimisation parameters need resolving in LT regression test
- #3654 LT optimisation-parameter awareness issue

@mkovari I'd like to get this merged independently; it only results in a <1% change in `te` and `dene` in the LT regression test. I'm happy to share other work I've done since on how this solves in different scenarios. Please approve when happy.

@timothy-nunn Please approve when satisfied with my changes/responses.

Thanks.

mkovari (approved):

timothy-nunn (approved):

timothy-nunn (approved):

PR #3652, opened 2025-05-01
Re-enable large tokamak (no f values)

timothy-nunn:
Was disabled in #3619. Since then, something in the code has changed to cause the solver to take a different path and solve, although I'm not sure what (which is unsurprising since we don't know what caused it, although #3619 was put in just after #3565).

### DIFF
--- a/tests/regression/test_process_input_files.py
+++ b/tests/regression/test_process_input_files.py
@@ -274,11 +274,6 @@ def test_input_file(
     should be compared in the test.
     :type opt_params_only: bool
     """
-    if input_file.name == "large_tokamak_nof.IN.DAT":
-        pytest.skip(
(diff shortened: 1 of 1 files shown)


mkovari:
Do you want to fix the error pointed out in #3574?
(`dr_tf_wp` and `dr_tf_inboard` are used as iteration variables simultaneously.)

jonmaddock (changes requested):
Great that this happens to converge again, but could you make Michael's changes about the mutually exclusive iteration variables? I'm aware this was a pre-existing problem, but it might be partly responsible for the convergence trouble in the first place.

timothy-nunn:
#3662 will now fix the issue with large tokamak nof

jonmaddock (approved):

PR #3653, opened 2025-05-02
More accurately simulate the neutron flux profile

OceanNuclear:
## Description
Calculates the neutron flux profile in the first wall and blanket, using an extremely simplified diffusion-based model of an infinite slab sandwich.

### Reason for choosing this model
It's better than not having a neutronics model.
It's also very simplified, so it can be calculated on the fly, without installing any deterministic neutron transport codes as dependency.

## Checklist


OceanNuclear:
Snapshot of where I'm at before taking a break for 3 weeks:

The current state of this branch is:
- [x] 1-group 2-layer model is written
- [x] 1-group 2-layer model is fully tested (regression & unit tests)
- [x] Multi-group model & >2-layer model is written
- [ ] Multi-group model & >2-layer model not yet tested
- [ ] Need to incorporate real data (The correct incident neutron flux from the plasma, and download the 1-group/multi-group cross-sections for various materials & then collapse them into the correct group structure.)
- [ ] Need to test the downloading and collapsing of cross-section.
- [x] Plotting is already available, example 1-group 2-layer model:
current
flux
heating

- [ ] I have no idea how to test plotting TBH.

je-cook (commented):
I'm not sure descriptors is the right way to go. Most descriptor stuff I've written I have eventually rewritten without descriptors because I found a better way.

On the atomic data is there any reason we cant use something like [periodictable](https://pypi.org/project/periodictable/) which we use in elsewhere.

OceanNuclear:
Listing all of the methods available under the NeutronProfile class (for my own use/sanity):

## Continuous range of values per layer:
| Quantity           | methods (manually select layer)      | methods (automatically select layer) | units           |
|--------------------|-----------------------------------------|-----------------------------------------|-----------------|
| Neutron flux       |`(groupwise_)neutron_flux_in_layer`      | `(groupwise_)neutron_flux_at`           | $m^{-2} s^{-1}$ | 
| Neutron current    |`(groupwise_)neutron_current_in_layer`   | `(groupwise_)neutron_current_at`        | $m^{-2} s^{-1}$ | 
| Heating            |`(groupwise_)neutron_heating_in_layer`   | `(groupwise_)neutron_heating_at`        | $W m^{-3}   $   | 
| Tritium production |`(groupwise_)tritium_production_in_layer`| `(groupwise_)tritium_production_at`     | mole $m^{-3}$   | 


## One value per layer:
| Quantity               | method(s)                                                                             | unit           |
|------------------------|---------------------------------------------------------------------------------------|----------------|
| Neutron flux           | `(groupwise_)integrated_flux_in_layer`                                                | $m^{-2} s^{-1}$      |
| Neutron current        | `(groupwise_)neutron_current_through_interface`/`(groupwise_)neutron_current_escaped` | $m^{-2} s^{-1}$      |
| Heating                | `(groupwise_)integrated_heating_in_layer`                                             | $W m^{-2}$         |
| Tritium production     | `(groupwise_)integrated_tritium_production_in_layer`                                  | mole $m^{-2} s^{-1}$ |
| linear heating density | `groupwise_linear_heating_density_in_layer`                                           | $J m^{-1}$         |

This gives a total of 27 available methods.

process/models/neutronics/base.py
```diff
@@ -0,0 +1,1580 @@
...
+        """Return number of coefficients pairs that has been populated."""
+        return self._len
+
+
+class AutoPopulatingDict:
+    """
+    Class that behaves like a dictionary, but if the required key does not
+    exist in the dictionary, it will call the populating_method to populate
+    that specific key.
+    """
```
je-cook, line 179: https://docs.python.org/3/library/collections.html#collections.defaultdict

PR #3667, opened 2025-05-20
Latest cmake installation docs

jonmaddock:
Instructions for installing latest cmake using pip, rather than older versions using apt.


clmould (approved):
Looks good to me!

PR #3687, opened 2025-06-12
Convert cost_variables to Python data structure

timothy-nunn:
# Rebase this PR rather than squash it

PR #3690, opened 2025-06-13
Water usage data structure

timothy-nunn:
Depends on #3687


clmould (changes requested):
Looks good, will approve and merge when conflicts are resolved

clmould (approved):

PR #3726, opened 2025-07-17
Convert CS fatigue data structure to Python

timothy-nunn:


process/data_structure/cs_fatigue_variables.py
```diff
@@ -0,0 +1,91 @@
...
+    n_cycle_min = 2.0e4
+
+    global t_structural_vertical
+    t_structural_vertical = 0.022
+
+    global t_structural_radial
+    t_structural_radial = 0.07
+
+    global bkt_life_csf
+    bkt_life_csf = 0
```
clmould, line 70: Can this be initialised to `0.0` as it is typed as a float not an int

process/data_structure/cs_fatigue_variables.py
```diff
@@ -0,0 +1,91 @@
+residual_sig_hoop: float = None
+"""residual hoop stress in strucutal material (Pa)"""
+
+n_cycle: float = None
+"""Allowable number of cycles for CS stress model"""
+
+n_cycle_min: float = None
+"""Minimum llowable number of cycles for CS stress model"""
```
clmould, line 8:
```suggestion
"""Minimum allowable number of cycles for CS stress model"""
```

clmould (changes requested):
Couple minor changes, good otherwise

process/data_structure/cs_fatigue_variables.py
clmould, line 70:
```suggestion
    bkt_life_csf = 0.0
```

clmould: Because it is typed as a float

clmould (approved):

PR #3786, opened 2025-08-13
Convert pfcoil data structure to Python

timothy-nunn:
Combines the `pfcoil.f90` and `pfcoil_variables.f90` into one data structure (still two init functions because the module variables need to be reinitialised every iteration)


timothy-nunn:
I have changed the test asset for `test_vsec`: `vsoh_exp = np.array(0.0)` --> `vsoh_exp = np.array(-407.276949)`. In the unit test `c_pf_coil_turn` is set as an array of shape (a, b, 1) which--for some reason--is not set across the interface properly and causes the array to be all 0's despite having non-zero values when its value is mocked. This does not happen in the Python data structure which causes a difference in the test. On main, I have `.squeeze`'d the `c_pf_coil_turn` array in the test and get the same failure (new value of -407.276949)

clmould (changes requested):
Some very minor changes

PR #3789, opened 2025-08-14
Separate example testing from integration suite

timothy-nunn:
Moves testing of the examples into a separate testing folder (`examples`). This means that both locally and on the CI is easy to run the integration suite without testing the example notebooks. This means the integration tests do not take as long now since they run in two halves.


.github/workflows/process.yml
```diff
@@ -90,6 +90,31 @@ jobs:
...
+      - name: Install PROCESS
+        # Editable install to match default install
+        run: |
+          pip install -e '.[test, examples]'
+          pip install -r requirements_dev.txt
+      - name: Install poppler
+        run: |
+          sudo apt update
+          sudo apt install poppler-utils
+      - name: Run integration tests
```
jonmaddock, line 115: Change to examples tests? (sorry)

jonmaddock (changes requested):
Great idea for speeding up tests.

PR #3794, opened 2025-08-15
Cache coolprop calculations to avoid calling PropsSI

timothy-nunn:
There has been a noticeable increase in the runtime of PROCESS recently. Inspecting the call graph, it shows that a majority of our time (46.63%) is spent in the TF coil `protect` method. Most of the time within this method (~39.25% of the total PROCESS runtime) is spent calculating fluid properties via the Coolprop library (`PropsSI` function).  

image

I have modified our interface to coolprop such that we cache the results from calling Coolprop's `PropsSI`. The new call graph now shows that only 9.53% of PROCESS' runtime is spent in the `protect` method of the TF coils. Calls to `PropsSI` are now not even recorded on the call graph their runtime is so small.

image

### DIFF
--- a/process/coolprop_interface.py
+++ b/process/coolprop_interface.py
@@ -1,54 +1,56 @@
+from functools import cache
+
 from CoolProp.CoolProp import PropsSI
 
 
 class FluidProperties:
     def __init__(self, coolprop_inputs: list[str | float]):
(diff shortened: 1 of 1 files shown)


jonmaddock (approved):
Well done for finding this: is that around a 30% speedup overall then?

PR #3828, opened 2025-09-08
Convert numerics to Python

timothy-nunn:
Converts the numerics data structure to Python. 

Adds a dummy Fortran file (`empty.f90`) with a barebones module structure to avoid 0-length lists in CMake. Otherwise, this PR would have to include the removal of CMake and f2py which would get long and complex quickly.


jonmaddock (approved):
Great!

PR #3856, opened 2025-09-19
Exempt np.linalg.solve call from nopython mode

timothy-nunn:
Thanks @nickodell for this great suggestion. This stops numba from compiling the `linalg.solve` against SciPy's solvers and uses numpy's linear algebra in normal (non-Numba) mode. I also took this opportunity to remove the `asfortranarray` because that is not necessary (and possibly has never been)


jonmaddock (changes requested):
Brilliant, but needs a comment about why this is required.

Issue #3027, opened 2024-01-25
`test_plane_stress` failures on certain OS


cjwgriesel:
FAILED tests/unit/test_sctfcoil.py::test_plane_stress[planestressparam0] - assert array([-2.442...61051384e+04]) == approx([0.0 ±...48 ± 3.6e-02])
FAILED tests/unit/test_sctfcoil.py::test_plane_stress[planestressparam1] - assert array([-2.442...61051384e+04]) == approx([0.0 ±...48 ± 3.6e-02])
================================================================= 2 failed, 426 passed, 1 skipped, 18 warnings in 1232.48s (0:20:32) ===

Additional info:-
>       assert sigr == pytest.approx(planestressparam.expected_sigr)
E       assert array([-2.442...61051384e+04]) == approx([0.0 ±...48 ± 3.6e-02])
E         comparison failed. Mismatched elements: 1 / 300:
E         Max absolute difference: 2.4424334654893232e-08
E         Max relative difference: 1.0
E         Index | Obtained                | Expected     
E         0     | -2.4424334654893232e-08 | 0.0 ± 1.0e-12

/home/ir-grie1/conda/Unruns/process_ADev1/PROCESS/tests/unit/test_sctfcoil.py:7683: AssertionError
========================================================================================== warnings summary ==========================================================================================
<frozen importlib._bootstrap>:241
  <frozen importlib._bootstrap>:241: RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 16 from C header, got 96 from PyObject

tests/integration/test_blanket_library_int.py::test_hydraulic_diameter
  /home/ir-grie1/conda/Unruns/process_ADev1/PROCESS/tests/conftest.py:124: UserWarning: 
          You are running the PROCESS test suite on an outdated system.
          This can cause floating point rounding errors in regression tests.

There is also a warning about running PROCESS on an outdated system so maybe that just needs updating????

From the output above when running the pytest on CSD3 all passes apart from two tests (test_plane_stress). Possibly might be fixed by updating the system? But not sure? Tried to update conda but stubbornly refused to?

timothy-nunn:
Also occurs on CSD3 login without using conda

timothy-nunn:
This issue also occurs on Mac: 

<img width="1124" height="544" alt="Image" src="https://github.com/user-attachments/assets/86d3f0d8-7f38-4674-9a41-df077d9a21d9" />

nickodell:
[...]

PR #3884, opened 2025-10-03
Scale linear system of equations in plane_stress and skip test if incompatible system

timothy-nunn:
Sadly, #3856 was short lived and enabling `numpy>=2` re-introduced the issues observed in #3027. 

I have been unable to find an acceptably fast solution that will ensure parity on different systems when solving $Ax=b$ in the `plane_stress` function. The issue arises (I believe) because the matrix `aa` is ill-conditioned (on the order of $10^{12}$). 

The first thing this PR introduces is a row-wise scaling of the `aa` matrix and `bb` vector such that the largest value on a row of `aa` will be `1.0`. This reduces the condition of `aa` significantly (to be on the order of $10^{3}$). However, one of the `plane_stress` tests still failed on Mac, the other two passed.

The second thing I have done is skip this test if the system is deemed incompatible (we actually already had a function for this that I wrote back in 2021). This means that the `plane_stress` tests are skipped on incompatible systems (e.g. Mac, Windows, CSD3) and will run when we think it should pass (on Ubuntu VMs and the CI). Maybe the reviewer could check that these tests don't get skipped on WSL/their Ubuntu laptop. 

I have also added a warning to the installation page to ensure this issue is abundantly obvious.

image

### DIFF
--- a/documentation/installation/installation.md
+++ b/documentation/installation/installation.md
@@ -14,6 +14,11 @@ PROCESS is a command-line Python program. As such, it should be supported by a w
     wsl --install
     ```
 
+!!! Warning "Users of 'incompatible' systems"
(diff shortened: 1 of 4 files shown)


timothy-nunn:
@chris-ashe checked on WSL and no unit tests are skipped

timothy-nunn:
Also skips on CSD3 (where #3027 was first observed)

Issue #3654, opened 2025-05-06
ST regression runs build model that is optimisation parameter-aware


jonmaddock:
Removal of this constraint causes constraint-aware code to not be executed hence the divide by 0 :/  https://github.com/ukaea/PROCESS/blob/b5572ebaa54827dca00b2170b7a31886ad326e28/process/build.py#L1689

_Originally posted by @timothy-nunn in https://github.com/ukaea/PROCESS/pull/3646#discussion_r2070405748_
            
The ST regression test runs part of the build model that is optimisation-parameter-aware; this is clearly bad practice and should be resolved. The test should be un-skipped when this is resolved.

PR #3894, opened 2025-10-10
Re-enable the ST evaluation regression test

timothy-nunn:
The ST evaluation regression file _appears_ to be running again without issue. It was disabled in #3646 because of divide by 0 errors and was set to be fixed in #3654. While the divide by 0 does not occur anymore and the file can be tested against, #3654 should still be resolved!


jonmaddock (approved):
I agree, if it's working again, then let's test it.

PR #3908, opened 2025-10-14
Improve quality of notebook renders

timothy-nunn:
Separates the logging setup into a new function so that it can be used in notebooks to suppress model logs to the terminal, as happens when process is run from the command line. 

Adds a magic command I noticed was missing so that the plot can render 

Removes certain metadata from being tracked in the notebooks (they would probably change each time someone edits the notebooks)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,4 +22,4 @@ repos:
     rev: 0.5.0
     hooks:
       - id: nbstripout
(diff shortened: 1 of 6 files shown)


clmould:
> The stuff in red is a `UserWarning` which is very similar to a log but is functionally different. This PR only concerns itself with removing logs. Although at some point it would be good practice to remove all of our user warnings and move them over to logs--this is a separate issue.

Okay great makes sense, I'll approve as it has removed the log stuff from the docs

Issue #3957, opened 2025-10-28
PROCESS scan functionality behaves unexpectedly


CoronelBuendia:
I have been performing a series of runs in PROCESS, for a typical conventional aspect ratio large tokamak problem, minimising R_0 for A in 2.6..3.6.

I used to do these kinds of runs frequently, prior to PROCESS being open-sourced. I would use the scan option liberally, usually just in one dimension. I found this feature useful, primarily because the run-time was better  (the previous solution in a scan is usually a good starting point for a subsequent point).

Often, the scan would fail at a given point, and I would just start a new scan with a different starting point and "approach" that failure from a different direction.

So I took it as a given when now starting to do such scans in PROCESS v3.2.1 that I should be scanning. However this fails (using `--solver vmcon_bounded`), even for moderate deviations from the starting solution.

In fact, for this particular scan, I can find solutions for 100 points using the _same_ starting vector (which I'm not sure would have worked in the last version of PROCESS I was using), but scanning I can't get much more than 12 or 13 points consecutively. I tried a range of different starting points and got these errors:

```
(/home/matti/miniforge/envs/bluemira/lib/python3.11/site-packages/process/tf_coil.py:246) dr_tf_plasma_case to small to accommodate the WP, forced to minimum value
Starting scan point 13 of 100 : Aspect_ratio , aspect = 3.03939394
/home/matti/miniforge/envs/bluemira/lib/python3.11/site-packages/cvxpy/problems/problem.py:1539: UserWarning: Solution may be inaccurate. Try another solver, adjusting the solver settings, or solve with verbose=True for more information.
  warnings.warn(
Traceback (most recent call last):-04
  File "/home/matti/miniforge/envs/bluemira/bin/process", line 7, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/matti/miniforge/envs/bluemira/lib/python3.11/site-packages/process/main.py", line 753, in main
    Process(args)
  File "/home/matti/miniforge/envs/bluemira/lib/python3.11/site-packages/process/main.py", line 127, in __init__
    self.run_mode()
  File "/home/matti/miniforge/envs/bluemira/lib/python3.11/site-packages/process/main.py", line 231, in run_mode
    self.run.run()
  File "/home/matti/miniforge/envs/bluemira/lib/python3.11/site-packages/process/main.py", line 390, in run
    self.run_scan()
[...]

PR #3964, opened 2025-10-30
Remove high relative tolerance on VMCON QPP

timothy-nunn:
This should go some of the way to improving the performance of VMCON, particularly when running a scan #3957. 

I noticed that when running a scan over the same point (e.g. `sweep = x, x, x, ..., x`) VMCON would perform many iterations on the non-first scan points. This does not make any sense because, after the first scan point, PROCESS is at a converged solution and VMCON should immediately identify this and not perform any iterations. I have verified that old VMCON did this as expected. 

VMCON's convergence criteria has two components:
- $|\nabla f(x)\cdot\delta|$: the predicted change in the objective function if VMCON is allowed another iteration.
- $\sum^m_{i=1}|\lambda_{i} c_i(x)|$: the complementary error.

In new VMCON, the first term was much higher than old VMCON, causing new VMCON to not believe it is converged and iterate in search of a solution. This was being driven by a difference in $\delta$. And the reason $\delta$ was different is because the relative tolerance on the quadratic programming solver was so high. 

With this PR:
- The $\delta$ calculated by new VMCON and old VMCON are near identical. 
- A PROCESS scan of the same point correctly does not iterate and converges immediately.


tests/integration/test_vmcon.py
```diff
@@ -44,13 +44,12 @@ def __init__(self, name, evaluator):
 
 
 class SolverArgs:
-    def __init__(self):
+    def __init__(self, n=2):
```
chris-ashe: Just a small comment. What is `n` in this case may be worth while updating the docstring.

chris-ashe (approved):

PR #3971, opened 2025-11-05
Use Clarabel instead of OSQP to solve QPP in VMCON

timothy-nunn:
I am noticing some non-determinisim in solver (specifically for the spherical tokamak regression test) where an "inaccurate solution" is returned from OSQP and this results in a non-deterministic step length. Clarbel was recommended for its determinism and accuracy (https://github.com/osqp/osqp-python/issues/116)

### DIFF
--- a/process/solver.py
+++ b/process/solver.py
@@ -4,6 +4,7 @@ import importlib
 import logging
 from abc import ABC, abstractmethod
 
+import cvxpy
 import numpy as np
 from pyvmcon import (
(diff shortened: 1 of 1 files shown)


timothy-nunn:
The use of Clarabel does not have an impact on convergence compared to current PROCESS:

These are the results of running 2000 input files within +/-30% of the starting position of the ST and LT-NOF regression tests (1000 from each test). 
image

Issue #3972, opened 2025-11-06
Different solutions on different systems


timothy-nunn:
Since the removal of the high tolerance in #3964, we have been noticing intermittent failures of the regression test in the CI on main. This is surprising because, on main, the regression tests for a commit `abcd` will use the assets just generated for that same commit `abcd`--there should be no difference.

I have explored whether this was non-determinism in the solver by changing to use the Clarabel quadratic programming problem solver #3971 however the problems persisted. That is not to say this was a bad decision and should be reversed, Clarabel is what we were recommended to use some years ago by the OSQP developers.

I have then checked whether the data going into and out of the QPP solve is _exactly_ the same. On my Mac, the arrays going into and out of the QPP in the `LT-NOF` case are exactly the same at all iterations, down to the last decimal place. However, when I compare my Mac and Linux VM, there are slight very small differences (attached below). This difference is present in the data coming from PROCESS **on the first iteration** and therefore does not result from VMCON/QPP solver. As the solver runs, this error accumulates by a couple of orders of magnitude. 

To verify that this is not new, I have gone back to a version of PROCESS before this change (58b2c96e17dcda22d99db7f6c835311522100327), downversioned to PyVMCON 2.3.1 and CVXPY 1.6. This is what the CI would have run using when it reliably passed. I then ran this test again between my Mac and Linux VM. The results show that there is indeed a difference between my Mac and Linux VM from the first iteration for the data coming out of PROCESS. The MAEs for the PROCESS data (`deq`, `ie`, `die`) on the first iteration are the same as the previous experiment. Unlikely in the previous experiments, the differences remain around the same order of magnitude and remains below the reporting threshold for PROCESS' test suite, even when running with `0%` tolerance. Indeed in our test suite we do not use an exact 0 tolerance for this reason 

https://github.com/ukaea/PROCESS/blob/8dd9e070983ba430c5ed2d9f7879aa8d3dae80c3/tests/regression/test_process_input_files.py#L192-L202

It is my opinion that reducing the QPP solver tolerance has just made apparent an issue which has existed in PROCESS all along. There are slight discrepancies between our results on different systems, but a lower QPP tolerance causes these to accumulate over a PROCESS run and results in very slightly different solver paths and hence solutions.

**Differences in `solve_qsp` data between Mac and Linux VM for a LT-NOF PROCESS run using Clarabel QPP solver**
```
deq@0 is different: MAE 1.6653345369377348e-13
[...]

PR #3973, opened 2025-11-06
Run 0.2% tolerance on CI/CD instead of 0%

timothy-nunn:
Temporary "fix" to the CI discussed in #3972


je-cook (approved):
As discussed offline, this seems the most sensible solution for now

PR #4031, opened 2026-01-05
Remove globals from plot proc and some cleanup

je-cook:
## Description

This does some fairly substantial cleanup of plot proc with the main aim of removing globals.

## Checklist

### DIFF
--- a/process/io/mfile.py
+++ b/process/io/mfile.py
@@ -209,11 +209,11 @@ class MFile:
             self.open_mfile()
             self.parse_mfile()
 
-    def get_variables(self, *variables: str, scan: int) -> list[Any]:
+    def get_variables(self, *variables: str, scan: int = -1) -> list[Any]:
         """Get a number of variables from a single scan"""
         return [self.get(v, scan=scan) for v in variables]
 
-    def get(self, variable: str, *, scan: int) -> Any:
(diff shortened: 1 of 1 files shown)


process/io/mfile.py
```diff
@@ -208,6 +209,12 @@ def __init__(self, filename="MFILE.DAT"):
             self.open_mfile()
             self.parse_mfile()
 
+    def get_variables(self, *variables: str, scan: int) -> list[Any]:
+        return [self.data[v].get_scan(scan) for v in variables]
+
+    def get(self, variable: str, scan: int) -> Any:
```
timothy-nunn:
```suggestion
    def get(self, variable: str, *, scan: int) -> Any:
```

In `get_variables`, scan is a keyword-only argument, I think this should be consistent between the two methods.

timothy-nunn (commented):
A couple of minor consistency and maintainability comments. Also a docstring for each of the new methods would be nice.

process/io/mfile.py
je-cook: ok, thoughts on having a default of -1 for scan?

timothy-nunn (approved):
Happy with changes in `process/io/mfile.py`

PR #4035, opened 2026-01-08
Write an additional log file in PROCESS working directory

timothy-nunn:
Writes an additional logfile in the PROCESS working directory (where the input file is) that follows the name of the input file:
- `my.IN.DAT` produces a `my.process.log` 
- `my_IN.DAT` produces a `my_process.log`

image

This does not remove the old `process.log` file which is created in the terminal working directory.

### DIFF
--- a/process/main.py
+++ b/process/main.py
@@ -111,6 +111,8 @@ from process.water_use import WaterUse
 
 os.environ["PYTHON_PROCESS_ROOT"] = os.path.join(os.path.dirname(__file__))
 
+PACKAGE_LOGGING = True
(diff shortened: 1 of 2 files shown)


je-cook (changes requested):
After thinking about what I said yesterday I agree with this change.

I think it would also be good to change the default file handler to append mode so you don't overwrite the log file.
change L727 to:
`logging_file_handler = logging.FileHandler("process.log", mode="a")`

alternatively you could rotate the logs somehow but that is more complex and possibly overkill (You could do a symlink process.log to a timed and dated filename)

PR #4038, opened 2026-01-08
Ensure i_plasma_pedestal=0 profile does not have 0 elements

timothy-nunn:
Adds in a check of the `profile_y` when `i_plasma_pedestal=0` to set any 0 elements to be slightly above 0, hence avoiding divide by 0's.


ym1906:
One more thing @timothy-nunn the change is only applied to the ne profile but actually the bug occurs with the te profile, we should apply the same solution there.

PR #4045, opened 2026-01-15
Unit test CI job not using matrix OS

timothy-nunn:
Despite running in a matrix job I have forgotten to use the matrix variable to change the OS


je-cook (approved):
:laughing:

PR #4105, opened 2026-02-23
Overhall CLI

### LINKED ISSUE #4011: :bug: `--mfile` arg does not work as expected
## Describe the bug

The `--mfile` argument when running PROCESS is not working as expected

## Steps to reproduce

`process -i path/to/IN.DAT --mfile TEST_MFILE.DAT`

## Expected behaviour

Running PROCESS as above should write the MFILE to your current working directory with the specified name (TEST_MFILE.DAT in this case), but it doesn't do this

## Environment

- OS: 
- Version:

je-cook:
## Description

Rewrites the CLI to use click and orchestrates all cli modules under the process cli function

Closes #4011
## Checklist

- [x] My changes follow the [PROCESS style guide](https://ukaea.github.io/PROCESS/development/standards/)


process/io/in_dat/cli.py
```diff
@@ -0,0 +1,28 @@
...
+    "-o",
+    "--indat-out",
+    "indat_out",
+    type=str,
+    default="new_IN.DAT",
+    help="IN.DAT to write out",
+)
+@click.option(
+    "-fpi",
+    "--feasuble-point-index",
```
timothy-nunn:
```suggestion
    "--feasible-point-index",
```

chris-ashe (commented):
Just a few requests.

- Can we add `-h` as a shortcut for `--help` as this was a legacy option?
- In some of the plotting options you have to specify `-f` before the MFILE's while in some the value is positional.
- In the Sankey plot section there is no option to plot the interactive `plotly` option. Only the old Matplotlib one is there.

timothy-nunn (commented):
Half way through my review and only comments so far are around filenames. Will post them now so they don't get lost over Easter.

timothy-nunn (changes requested):
The `--mfile` fix is only half working. If I run:
`process -i large_tokamak_nof.IN.DAT --mfile my.MFILE.DAT --full-output` then it fails because it looks for `my.SIG_TF.json`... I would suggest propagating the mfile prefix to the OUT.DAT, SIG_TF.json, and the .log file.

timothy-nunn (commented):
Still also getting errors for the two commands not ticked in https://github.com/ukaea/PROCESS/pull/4105#issuecomment-4170709146

timothy-nunn (approved):

PR #4122, opened 2026-03-05
Use local copy of regression data for regression tests

timothy-nunn:
Moves from getting regression reference data using `requests` to using the local git client to get the regression references. This does require the use of a FileLock because when using the `pytest-xdist` plugin multiple threads start trying to modify the same directory (by doing `git fetch` and `git pull`)


je-cook (commented):
Seems reasonable. Just a couple of questions

PR #4164, opened 2026-04-07
4163 inconsistent bosch hale correction in beam fusion calculation

### LINKED ISSUE #4163: Inconsistent Bosch-Hale correction in beam fusion calculation
The beam fusion model is used to calculation fusion power from beam/plasma interactions.

`beam_fusion()` -> `beamcalc()` -> `alpha_power_beam()` -> `bosch_hale_reactivity()`

`bosch_hale_reactivity` should be passed a profile, currently it is passed a volume averaged temperature. 

This results in a much higher value of beam_fusion being calculated than might be expected and is inconsistent with the way the function should work according to its own doc-string and reference..

This is also inconsistent with how the calculation is done elsewhere in the code, where `fusion_rate_integral()` is used.

This effects a correction ratio for the reactivity in `alpha_power_beam()`  and has a significant impact on the beam fusion power calculated by PROCESS.

To amend this, we should pass in an ion temperature profile directly; and integrate this as is done elsewhere in the code.

grmtrkngtn:
## Description

The beam fusion model is used to calculation fusion power from beam/plasma interactions.

beam_fusion() -> beamcalc() -> alpha_power_beam() -> bosch_hale_reactivity()

bosch_hale_reactivity should be passed a profile, currently it is passed a volume averaged temperature.

This results in a much higher value of beam_fusion being calculated than might be expected and is inconsistent with the way the function should work according to its own doc-string and reference..

This is also inconsistent with how the calculation is done elsewhere in the code, where fusion_rate_integral() is used.

This affects a correction ratio for the reactivity in alpha_power_beam() and has a significant impact on the beam fusion power calculated by PROCESS.

To amend this, we should pass in an ion temperature profile directly; and integrate this as is done elsewhere in the code, and use this integral to calculate the correction ratio.

With this change PROCESS now produces a beam fusion power in line with another code - METIS - for a machine of similar size and operation.

## Checklist


process/models/physics/fusion_reactions.py
```diff
@@ -992,22 +990,57 @@ def beam_fusion(
...
+        ),
+        x=plasma_profile.neprofile.profile_x,
+        dx=plasma_profile.neprofile.profile_dx,
+    )
+
+    # Correction factor between the reference profile-averaged DT reactivity
+    # and the Bosch-Hale reactivity evaluated with the same profile integral.
+    sigmav_dt_correction = sigmav_dt_average / bh_profile_average
+
+    hot_deuterium_rate = 1e-4 * beam_reaction_rate(
```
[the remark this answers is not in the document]
chris-ashe: Ah ok, Can I ask that the variable name is changed to represent what it really is. Looking at the docstring of `beam_reaction_rate` it says it gives out `float: Hot beam fusion reaction rate (m^3/s).`. In this case the `fusden_` prefix should be used. Also if it actually in per cm cubed then please add `_cm` to the end of the name

process/models/physics/fusion_reactions.py
```diff
@@ -992,22 +992,57 @@ def beam_fusion(
...
+        constants.M_DEUTERON_AMU, deuterium_critical_energy_speed, e_beam_kev
+    )
+
+    hot_tritium_rate = 1e-4 * beam_reaction_rate(
+        constants.M_TRITON_AMU, tritium_critical_energy_speed, e_beam_kev
+    )
+
+    deuterium_beam_alpha_power = alpha_power_beam(
```
chris-ashe: As per style guide. `p_beam_deuterium_dt` is the recommended name`

process/models/physics/fusion_reactions.py
```diff
@@ -992,22 +992,57 @@ def beam_fusion(
...
+        deuterium_beam_density,
+        tritium_density,
+        hot_deuterium_rate,
+        vol_plasma,
+        sigmav_dt_correction,
+    )
+
+    tritium_beam_alpha_power = alpha_power_beam(
```
chris-ashe: Same as above

chris-ashe (changes requested):
Looks good logically. Are we not adding the profile correction for the beam width? Should be easy enough to do.

The docs also need to be updated.

I am not asking to replace everything, but can new variables you add please follow the style guide

process/models/physics/fusion_reactions.py
grmtrkngtn: Resolved.

process/models/physics/fusion_reactions.py
grmtrkngtn: Fixed

grmtrkngtn:
> Looks good logically. Are we not adding the profile correction for the beam width? Should be easy enough to do.
> 

I would prefer to keep this PR for the correction + refactor, then make a separate issue for the beam width (new phyiscs)

process/models/physics/fusion_reactions.py
grmtrkngtn: i changed hot_deuterium_rate to sigv_beam_deuterium. It's not a fusion density so I don't think fusden is right.

process/models/physics/fusion_reactions.py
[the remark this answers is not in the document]
grmtrkngtn: I've moved this conversion into the beam_reaction_rate function. I think this is better as it now outputs SI units. Also updated the docstring.

process/models/physics/fusion_reactions.py
grmtrkngtn: I changed quite a few additional variable names

grmtrkngtn:
> The docs also need to be updated.


Updated the docs now

process/models/physics/fusion_reactions.py
```diff
@@ -988,106 +988,132 @@ def beam_fusion(
...
-        range of beam velocities up to the critical velocity.
+        - Beam ion densities are derived from the beam source rate and slowing-down time.
+        - Critical energies define the transition between electron- and ion-dominated slowing down.
+        - Critical speeds are computed from the corresponding critical energies.
+        - The deposited energy represents the average energy retained by beam ions during slowing down.
 
     References:
-        - H.-S. Bosch and G. M. Hale, “Improved formulas for fusion cross-sections and thermal reactivities,”
+        - H.-S. Bosch and G. M. Hale, “Improved formulas for fusion cross-sections and thermal reactivities,”Collapse commentComment on line L1090chris-ashe commented on Apr 9, 2026 chris-asheon Apr 9, 2026CollaboratorMore actionsWhy have references been removedReactAdd diff commentMarkdown input:  edit mode selected.WritePreviewHeadingBoldItalicQuoteCodeLinkUnordered listNumbered listTask listMentionReferenceSaved repliesMistake, I've added them back in.Add FilesPaste, drop, or click to add filesCancelReplyStart a reviewResolve comment
```
chris-ashe: Copy paste error

chris-ashe (changes requested):
Image

Docs are not rendering properly. I think its because you have put heading inside indented lists

timothy-nunn (commented):
Gone through the code and happy with quality (docstrings, typing, testing, etc) so will be happy once Chris is.

process/models/physics/fusion_reactions.py
grmtrkngtn: oops, fixed now.

grmtrkngtn:
> Image
> 
> Docs are not rendering properly. I think its because you have put heading inside indented lists

Fixed

grmtrkngtn:
I'm going to close this PR because the issue is slightly different than I thought it was, and now has a different solution. The re-factoring work will carry on to the new issue/pr. 

I don't think Bosch-Hale should be used in beam_fusion() calculations at all, the new issue is to remove it.

@chris-ashe @timothy-nunn

PR #4184, opened 2026-04-14
Strict ruff linting

je-cook:
## Description

Makes the linting stricter. Adds a new linting job post the required one that has failures we'd like to fix but there are a lot so hopefully it will encourage people to fix the lints that make the CI fail for their PRs

## Checklist

### DIFF
--- a/.github/workflows/process.yml
+++ b/.github/workflows/process.yml
@@ -116,6 +116,27 @@ jobs:
       - name: Run Hatch linting tasks
         uses: Fusion-Power-Plant-Framework/fppf-actions/hatch-lint@c6af22fc8d00be67cca53a61177301462f499fb2
 
+  improve_quality_check:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    needs: [pre-commit-quality-check]
+    steps:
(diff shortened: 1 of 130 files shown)


.github/workflows/process.yml
```diff
@@ -116,6 +116,46 @@ jobs:
...
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
+
+      - name: Run improvement lint
+        if: steps.changes.outputs.files != ''
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          mapfile -t files <<< "${{ steps.changes.outputs.files }}"
```
timothy-nunn: My understanding of shell scripts is probably a bit rusty. ... But this line seems to set a `files` variable (array) with the changes files. I am failing to see what the point is of the step above then since a `files` variable is created there but is not used in this step?

je-cook: So if you pass the raw string directly to ruff it didnt seem to work but looking at it lets see if it does work.

je-cook: See https://github.com/ukaea/PROCESS/actions/runs/24449882923/job/71435847434 for failure if that is not done

timothy-nunn (approved):

PR #4187, opened 2026-04-14
Fix nbi power: correct shine-through, use plasma-coupled power instead of raw power; corrected wall-plug power conversion

### LINKED ISSUE #4186: Fix neutral beam shine-through and power accounting bugs
There are a few issues to fix in the NBI power accounting:

- Shine-through power incorrectly calculated
     - Shine-through fraction applied to only part of the power term due to operator precedence
     - Shine-through fraction applied incorrectly to work out beam shine-through
- Ion/electron power split uses wrong base power
    - Split performed on injected beam power instead of plasma-coupled beam power
- Wall-plug power (pwpnb) incorrectly computed
    - Only part of the power term divided by injector efficiency due to missing parentheses

all in current_drive.py

grmtrkngtn:
## Description

- We were applying f_p_beam_shine_through incorrectly in two ways:
     - `f_p_beam_shine_through` represents the fraction of beam power that is not absorbed, but we were using `(1 - f_p_beam_shine_through)`
     -  the fraction was only being applied to part of the power term due to operator precedence, rather than the total beam power
- Fixed ion/electron power split:
    - previously based on injected beam power
    - now correctly based on plasma-coupled beam power
- Fixed wall-plug power calculation (pwpnb):
    - previously only part of the power term was divided by injector efficiency
    - now the full beam power term is divided by efficiency

## Checklist


process/models/physics/current_drive.py
kj5248, line 188: I know you didn't add this but I think something here is incorrect. This should be unitless as it provides the fraction of input beam that reaches target location. From what I can gather the dpath is in m, nd_plasma_electrons_vol_avg is in m^-3 and sigbeam says it outputs in m^-2 which is not unitless. I think it is just a mistake with the sigbeam function as normally microscopic cross section is in units of m^2 but worth checking.

kj5248 (changes requested):
Looked through it and I think the changes to how we use f_p_beam_shine_through is definitely correct as is switching to using the deposited beam power for the ion/electron power split calculation. Good catch with the wall-plug power calculation someone definitely stuck the brackets in the wrong place. If you double check the sigbeam function and probably just change the function description and push that I will approve it.

process/models/physics/current_drive.py
grmtrkngtn: Good catch, I think you're right and it's just a typo in the docstring. I've correct this in the latest commit.

je-cook (approved):
I'll merge once the conflicts are fixed

je-cook (approved):

PR #4194, opened 2026-04-16
Add supercon tf wp shape types

chris-ashe:
## Checklist

### DIFF
--- a/process/core/init.py
+++ b/process/core/init.py
@@ -65,6 +65,7 @@ from process.models.superconductors import (
     SuperconductorType,
 )
 from process.models.tfcoil.base import TFCoilShapeModel, TFConductorModel
+from process.models.tfcoil.superconducting import SuperconductingTFWPShapeType
 
 if TYPE_CHECKING:
     from process.core.model import DataStructure
@@ -1002,9 +1003,13 @@ def check_process(inputs, data):  # noqa: ARG001
 
     if data_structure.tfcoil_variables.i_tf_wp_geom == -1:
         if data_structure.tfcoil_variables.i_tf_turns_integer == 0:
-            data_structure.tfcoil_variables.i_tf_wp_geom = 1
+            data_structure.tfcoil_variables.i_tf_wp_geom = (
+                SuperconductingTFWPShapeType.DOUBLE_RECTANGULAR
+            )
         if data_structure.tfcoil_variables.i_tf_turns_integer == 1:
-            data_structure.tfcoil_variables.i_tf_wp_geom = 0
+            data_structure.tfcoil_variables.i_tf_wp_geom = (
+                SuperconductingTFWPShapeType.RECTANGULAR
+            )
 
     # Setting the TF coil conductor elastic properties
 
@@ -1102,7 +1107,8 @@ def check_process(inputs, data):  # noqa: ARG001
         )
 
     if (
-        data_structure.tfcoil_variables.i_tf_wp_geom != 0
+        data_structure.tfcoil_variables.i_tf_wp_geom
+        != SuperconductingTFWPShapeType.RECTANGULAR
         and data_structure.tfcoil_variables.i_tf_turns_integer == 1
     ):
         raise ProcessValidationError(
--- a/process/core/init.py
+++ b/process/core/init.py
@@ -1000,8 +1000,10 @@ def check_process(inputs, data):  # noqa: ARG001
             data_structure.tfcoil_variables.eyoung_ins = 2.5e9
 
     # Setting the default WP geometry
-
-    if data_structure.tfcoil_variables.i_tf_wp_geom == -1:
+    i_tf_wp_geom = SuperconductingTFWPShapeType(
+        data_structure.tfcoil_variables.i_tf_wp_geom
+    )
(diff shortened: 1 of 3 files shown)


process/models/tfcoil/superconducting.py
```diff
@@ -48,6 +48,25 @@ def abbreviation(self):
...
+    def __new__(cls, value, full_name):
+        obj = int.__new__(cls, value)
+        obj._value_ = value
+        obj._full_name_ = full_name
+        return obj
+
+    @DynamicClassAttribute
+    def full_name(self):
+        """Return the full name for this WP geometry type."""
+        return self._full_name_
```
je-cook: Then you dont need a `__new__` at all and it can be `RECTANGULAR = 0`, it works on all of the above entries I checked.
```suggestion
        return self.name.title().replace("_", " ")
```

process/models/tfcoil/superconducting.py
```diff
@@ -1020,7 +1039,7 @@ def superconducting_tf_wp_geometry(
```
je-cook, line 1034: Need a line of validation here:

```suggestion
i_tf_wp_geom = SuperconductingTFWPShapeType(i_tf_wp_geom)
```

process/models/build.py
```diff
@@ -1579,15 +1580,20 @@ def plasma_outboard_edge_toroidal_ripple(
             r_wp_min = r_tf_wp_inboard_inner
```
je-cook, line 1581: and a validation line here

process/models/tfcoil/superconducting.py
```diff
@@ -48,6 +48,25 @@ def abbreviation(self):
...
+
+    RECTANGULAR = (0, "Rectangular")
+    DOUBLE_RECTANGULAR = (1, "Double Rectangular")
+    TRAPEZOIDAL = (2, "Trapezoidal")
+
+    def __new__(cls, value, full_name):
+        obj = int.__new__(cls, value)
+        obj._value_ = value
+        obj._full_name_ = full_name
+        return obj
```
je-cook: As I said before this isnt needed
```suggestion
    RECTANGULAR = 0
    DOUBLE_RECTANGULAR = 1
    TRAPEZOIDAL = 2
```

je-cook (changes requested):
Please also do my previous review comments

process/models/tfcoil/superconducting.py
je-cook: you need to tab this in...

process/models/tfcoil/superconducting.py
```diff
@@ -48,6 +48,20 @@ def abbreviation(self):
         return self._abbreviation_
 
 
+class SuperconductingTFWPShapeType(IntEnum):
+    """Enum for the type of TF coil WP shape, which determines the geometry of the winding pack and ground insulation."""
+
+    DEFAULT = -1
```
je-cook: What is the default? If I'm being really picky they all should have docstrings but I'd take a definition of default or at least pointing to it in the docs.

chris-ashe: Its found in the initialisation. Was causing the tests to fail. It may be best to change this for when we add the integer turns `enum`. Only integer turns can be used with the rectangular WP shape. Hence the logic below.

```python
 # Setting the default WP geometry

    if data_structure.tfcoil_variables.i_tf_wp_geom == -1:
        if data_structure.tfcoil_variables.i_tf_turns_integer == 0:
            data_structure.tfcoil_variables.i_tf_wp_geom = 1
        if data_structure.tfcoil_variables.i_tf_turns_integer == 1:
            data_structure.tfcoil_variables.i_tf_wp_geom = 0
```

je-cook: probably then DEFAULT should be changed to UNSET maybe? then its all pretty explanatory names unless I'm reading it wrong?

chris-ashe: The whole block is mainly to prevent error raising down the line. If the turns are integer it sets it to the only working model which is the rectangular shape. If they arent integer it will default to double rectangular

je-cook: That is UNSET not default then as the default is rectangular. 3 options:

1) add a `def _missing_` to the enum eg https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/2cdfa8265abda413456a401b6df4dd79efa726ef/bluemira/fuel_cycle/analysis.py#L41 . This can then send -1 -> rectangular
2) change the logic so that -1 is never put into the enum
3) call it unset not default as thats a misnomer

I would say 3 is the simplest

process/core/init.py
```diff
@@ -1002,9 +1003,13 @@ def check_process(inputs, data):  # noqa: ARG001
 
     if data_structure.tfcoil_variables.i_tf_wp_geom == -1:
```
je-cook:
```suggestion
    i_tf_wp_geom = SuperconductingTFWPShapeType(data_structure.tfcoil_variables.i_tf_wp_geom)
 
    if i_tf_wp_geom == SuperconductingTFWPShapeType.UNSET:
```

je-cook (approved):

PR #4206, opened 2026-04-27
varyrun stop IN.DAT being created

### LINKED ISSUE #4204: VaryRun file creation
VaryRun creates IN.DAT, MFILE.DAT and OUTFILE.DAT as well as 0_IN.DAT, 0_MFILE.DAT and 0_OUT.DAT. These files without the numbered prefix should not be created. IN.DAT was being created at end of `check_in_dat` in vary_run/tools.py.
Update varyrun example accordingly

clmould:
## Description

Closes #4204 
Stops creating IN.DAT, MFILE.DAT and OUT.DAT in VaryRun runs (so now only makes the 0_, 1_ etc), update the `__next__` method so no files are created once the max iterations are reached (as before if it reached max iterations it would create the next IN.DAT then error saying it can't find the associated MFILE), add messages to make clear if that iteration's file converged or not, update docs and varyrun example

## Checklist


timothy-nunn (changes requested):
The old files are now not being deleted:

Image

We only want the `2_*` files to be left (in this example) as the others a produced by non-convergent input files

clmould:
> The old files are now not being deleted:
> 
> Image
> We only want the `2_*` files to be left (in this example) as the others a produced by non-convergent input files

@timothy-nunn This is how it's currently working on main

timothy-nunn:
> > The old files are now not being deleted:
> > Image
> > We only want the `2_*` files to be left (in this example) as the others a produced by non-convergent input files
> 
> @timothy-nunn This is how it's currently working on main

@clmould the image is not working for me, but that surprises me. I don't think thats how it worked when I last used vary run, and I don't think that is how it should work because the intermediate files are not useful imo. This PR seems like a good opportunity to fix this.

clmould:
Keeping the intermediate files for now as removing them wasn't as straightforward as initially thought. #4227

process/core/io/vary_run/config.py
```diff
@@ -212,20 +230,33 @@ def create_readme(self):
...
-        error_status = (
-            f"Error status: {m_file.data['error_status'].get_scan(-1)}  "
-            f"Error ID: {m_file.data['error_id'].get_scan(-1)}\n"
-        )
+        try:
+            error_status = (
+                f"Error status: {m_file.data['error_status'].get_scan(-1)}  "
+                f"Error ID: {m_file.data['error_id'].get_scan(-1)}\n"
+            )
+        except:  # noqa: E722
```
timothy-nunn: Please `except KeyError` here (I expect this is the error to be caught here)

process/core/io/vary_run/config.py
```diff
@@ -212,20 +230,33 @@ def create_readme(self):
...
+            error_status = ""
```
timothy-nunn: If an error occurs in the above line, I expect this means the MFILE is completely empty and you could probably state that here

process/core/io/vary_run/config.py
```diff
@@ -494,15 +530,16 @@ def modify_ixc(self):
             )
             sys.exit()
 
-        in_dat = InDat(filename="0_IN.DAT")
+        # Fine here
```
timothy-nunn:
```suggestion

```

timothy-nunn (changes requested):

timothy-nunn (approved):

examples/vary_run_example.ex.py
```diff
@@ -120,7 +120,7 @@
 # VaryRun always produces a file called IN.DAT in the same directory
 # as the conf file
 _, updated_iteration_variable_values = get_mfile_initial_ixc_values(
-    Path(temp_dir.name) / "IN.DAT", vary_run.data
+    Path(temp_dir.name) / "2_IN.DAT", vary_run.data
```
timothy-nunn, line 123: Hardcoding that it always finishes on `2` might be problematic and cause the example to be wrong or crash. Is there a way of finding the last file VaryRun produced?

timothy-nunn (changes requested):

examples/vary_run_example.ex.py
je-cook: atm not straightforwardly. It is stored on the RunProcessConfig object but that is not stored as an attribute on vary run IIRC

timothy-nunn: With a re-write of VaryRun hopefully coming soon (™️) it might be worth just leaving this but being mindful then ...?

timothy-nunn (approved):

PR #4225, opened 2026-05-06
Remove commented out code

timothy-nunn:
Found some commented out code, mainly old format statements from the Fortran days

### DIFF
--- a/process/core/io/plot/sankey.py
+++ b/process/core/io/plot/sankey.py
@@ -880,7 +880,6 @@ def plot_sankey(
 
     # Munipulating the positioning of the branch labels
     # -ve to left and down; +ve to right and up
-    # pos[0] = x-axis; pos[1] = y-axis
     for d in diagrams:
         for y, t in enumerate(d.texts):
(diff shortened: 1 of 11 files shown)


process/models/power.py
```diff
@@ -2442,8 +2442,6 @@ def tfcpwr(
                 "OP ",
             )
             # MDK Remove this as it leads to confusion between (a) total inductance/n_tf_coils, or (b)
```
clmould: I think this needs deleting

process/core/io/plot/sankey.py
```diff
@@ -880,7 +880,6 @@ def plot_sankey(
 
     # Munipulating the positioning of the branch labels
```
clmould: Not sure if this might need deleting? If not, could the spelling be fixed just while you're here

clmould (changes requested):

process/core/io/plot/sankey.py
timothy-nunn: Think the comment is correct, have corrected spelling

clmould (approved):

PR #4230, opened 2026-05-07
Fwbs dataclass

clmould:
## Checklist

### DIFF
--- a/process/core/solver/iteration_variables.py
+++ b/process/core/solver/iteration_variables.py
@@ -1,5 +1,6 @@
 from copy import deepcopy
 from dataclasses import dataclass
+from typing import Any
 from warnings import warn
 
 import numpy as np
@@ -13,7 +14,7 @@ from process.core.model import DataStructure
 class IterationVariable:
     name: str
     """The name of the variable"""
-    module: str
+    module: str | Any
     """The Fortran module that this variable should be set on."""
     lower_bound: float
     """The default lower bound of the iteration variable"""
@@ -322,18 +323,21 @@ def load_iteration_variables(data):
         iteration_variable = ITERATION_VARIABLES[variable_index]
 
         # use ... as the default return value because None might be a valid return from Fortran?
+
+        module = (
+            getattr(data, iteration_variable.module)
+            if isinstance(iteration_variable.module, str)
+            else iteration_variable.module
+        )
+
         iteration_variable_value = getattr(
-            iteration_variable.module,
+            module,
             iteration_variable.target_name or iteration_variable.name,
             ...,
         )
-        if isinstance(iteration_variable.module, str):
-            iteration_variable_value = getattr(
-                getattr(data, iteration_variable.module),
-                iteration_variable.target_name or iteration_variable.name,
-            )
 
-        elif iteration_variable_value is ...:
+        # If iteration variable is missing
+        if iteration_variable_value is ...:
             error_msg = (
                 f"Could not get the value for iteration variable {variable_index} "
                 f"({iteration_variable.name})"
@@ -404,40 +408,28 @@ def set_scaled_iteration_variable(xc, nn: int, data: DataStructure):
 
         ratio = xc[i] / data_structure.numerics.scale[i]
 
+        module = (
+            getattr(data, iteration_variable.module)
+            if isinstance(iteration_variable.module, str)
+            else iteration_variable.module
+        )
+
         if iteration_variable.array_index is None:
-            if isinstance(iteration_variable.module, str):
-                setattr(
-                    getattr(data, iteration_variable.module),
-                    iteration_variable.target_name or iteration_variable.name,
-                    ratio,
-                )
-            else:
-                setattr(
-                    iteration_variable.module,
-                    iteration_variable.target_name or iteration_variable.name,
-                    ratio,
-                )
-        elif isinstance(iteration_variable.module, str):
-            current_array = getattr(
-                getattr(data, iteration_variable.module),
-                iteration_variable.target_name or iteration_variable.name,
-            )
-            new_array = deepcopy(current_array)
-            new_array[iteration_variable.array_index] = ratio
             setattr(
-                getattr(data, iteration_variable.module),
+                module,
                 iteration_variable.target_name or iteration_variable.name,
-                new_array,
+                ratio,
             )
+
         else:
             current_array = getattr(
-                iteration_variable.module,
+                module,
                 iteration_variable.target_name or iteration_variable.name,
             )
             new_array = deepcopy(current_array)
             new_array[iteration_variable.array_index] = ratio
             setattr(
-                iteration_variable.module,
+                module,
                 iteration_variable.target_name or iteration_variable.name,
                 new_array,
             )
--- a/process/core/solver/iteration_variables.py
+++ b/process/core/solver/iteration_variables.py
@@ -1,19 +1,19 @@
 from copy import deepcopy
 from dataclasses import dataclass
-from typing import Any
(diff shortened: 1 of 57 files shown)


process/core/solver/iteration_variables.py
```diff
@@ -1,19 +1,19 @@
...
 from process.core.exceptions import ProcessValueError
+from process.core.model import DataStructure
 
 
 @dataclass
 class IterationVariable:
     name: str
     """The name of the variable"""
-    module: Any
+    module: str
```
timothy-nunn: This type annotation is wrong. The correct one at the moment is probably `str | Any`

process/core/solver/iteration_variables.py
```diff
@@ -331,8 +327,13 @@ def load_iteration_variables():
...
             ...,
         )
+        if isinstance(iteration_variable.module, str):
+            iteration_variable_value = getattr(
+                getattr(data, iteration_variable.module),
+                iteration_variable.target_name or iteration_variable.name,
+            )
 
-        if iteration_variable_value is ...:
+        elif iteration_variable_value is ...:
```
timothy-nunn: I think the logic here could be simplified. 

if module is string: get the data from the data structure
else: get the data from the module

timothy-nunn: This shouldn't be `elif` because we want to check `iteration_variable_value` regardless of where it has come from

process/core/solver/iteration_variables.py
```diff
@@ -331,8 +327,13 @@ def load_iteration_variables():
             iteration_variable.target_name or iteration_variable.name,
             ...,
```
timothy-nunn, line 336: Notice this default value, the same default needs to be returned from the getattr below

process/core/solver/iteration_variables.py
```diff
@@ -403,10 +405,29 @@ def set_scaled_iteration_variable(xc, nn: int):
...
+                    getattr(data, iteration_variable.module),
+                    iteration_variable.target_name or iteration_variable.name,
+                    ratio,
+                )
+            else:
+                setattr(
+                    iteration_variable.module,
+                    iteration_variable.target_name or iteration_variable.name,
+                    ratio,
+                )
```
timothy-nunn: There is a lot of duplication here which can be simplified 

```python
module = ... if isinstance(str, iteration_variable.module) else ...
setattr(module, ...)
```

process/core/solver/iteration_variables.py
timothy-nunn: Again, a lot of duplication here that I think could be reduced by not adding an additional `elif` statement and just having a `module` variable

process/data_structure/blanket_library.py
```diff
@@ -121,14 +121,6 @@
...
 
 
-n_fw_inboard_channels: float = None
-"""Inboard total number of first wall coolant channels"""
-
-
-n_fw_outboard_channels: float = None
-"""Outboard total number of first wall coolant channels"""
-
-
```
timothy-nunn, line 131: Why are these being moved?

process/models/engineering/materials.py
```diff
@@ -40,6 +38,6 @@ def eurofer97_thermal_conductivity(temp: float) -> float:
     # temp in Kelvin
     return (
         (5.4308 + 0.13565 * temp - 0.00023862 * temp**2 + 1.3393e-7 * temp**3)
-        * fwbs_variables.fw_th_conductivity
+        * data.fwbs.fw_th_conductivity
```
timothy-nunn: I'm not a fan of passing the whole data structure into this function. The function signature should be:

```python
def eurofer97_thermal_conductivity(temp: float, fw_th_conductivity: float) -> float:
```

and the function body/where its called should be updated accordingly

process/models/ivc_tools.py
timothy-nunn, line 1: I'm a little confused why this file has been created?

timothy-nunn (changes requested):
I haven't quite checked everything but here are my comments on the main logic changes

process/models/ivc_tools.py
clmould: There was a circular import between the first wall and blanket, and this removed that. Also these functions aren't specific to the blanket and are used by other modules so made sense to remove them from blanket_library

clmould: I'll move this to another PR

process/data_structure/blanket_library.py
clmould: Moved back now

clmould:
Since this is the first PR to required iteration_variables to be changes, I tested out using some fwbs iteration variables in large_tokamak_IN.DAT:
image

I then ran process on this file on this branch and also on main:
image

The only difference was in runtime.

process/core/solver/iteration_variables.py
```diff
@@ -403,20 +409,34 @@ def set_scaled_iteration_variable(xc, nn: int):
...
```
timothy-nunn: This logic here can also be simplified by introducing a `module` variable, similar to what you did above and below.

process/data_structure/blanket_variables.py
```diff
@@ -0,0 +1,187 @@
...
+    """Blanket internal half-height (m)"""
+
+    dz_shld_half: float = 0.0
+    """Shield internal half-height (m)"""
+
+    dz_vv_half: float = 0.0
+    """Vacuum vessel internal half-height (m)"""
+
+    icomponent: int = None
+    """Switch used to specify selected component: blanket=0, shield=1, vacuum vessel=2"""
```
timothy-nunn: Can be removed

process/models/power.py
```diff
@@ -1618,27 +1617,27 @@ def plant_electric_production(self):
 
         #  Total secondary heat
         #  (total low-grade heat rejected - does not contribute to power conversion cycle)
-        #  Included fwbs_variables.p_tf_nuclear_heat_mw
-        # p_plant_secondary_heat_mw = power_variables.p_plant_core_systems_elec_mw + heat_transport_variables.p_hcd_electric_loss_mw + heat_transport_variables.p_coolant_pump_loss_total_mw + hthermmw + heat_transport_variables.p_div_secondary_heat_mw + heat_transport_variables.p_shld_secondary_heat_mw + heat_transport_variables.p_hcd_secondary_heat_mw + fwbs_variables.p_tf_nuclear_heat_mw
+        #  Included self.data.fwbs.p_tf_nuclear_heat_mw
+        # p_plant_secondary_heat_mw = power_variables.p_plant_core_systems_elec_mw + heat_transport_variables.p_hcd_electric_loss_mw + heat_transport_variables.p_coolant_pump_loss_total_mw + hthermmw + heat_transport_variables.p_div_secondary_heat_mw + heat_transport_variables.p_shld_secondary_heat_mw + heat_transport_variables.p_hcd_secondary_heat_mw + self.data.fwbs.p_tf_nuclear_heat_mw
```
timothy-nunn:
```suggestion
```
Whilst we are editing this, this commented code can go

process/data_structure/blanket_variables.py
timothy-nunn, line 1: I see this file is created, but has the old one been removed?

timothy-nunn (changes requested):

process/data_structure/blanket_variables.py
clmould: Mistakenly pushed

process/data_structure/blanket_variables.py
clmould: I'll do this in the blanket variable dataclass PR

timothy-nunn (approved):

PR #4235, opened 2026-05-08
Remove circular FW and blanket, and create ivc_tools.py for common fns

clmould:
## Description

Removes circular imports of blanket and first wall
Moves functions common to IVCs into separate file, instead of being within the blanket_library.py:
`calculate_pipe_bend_radius`, `dshellarea`, `dshellvol`, `eshellarea`, `eshellvol`

## Checklist

### DIFF
--- a/process/models/engineering/ivc_functions.py
+++ b/process/models/engineering/ivc_functions.py
@@ -24,7 +24,7 @@ def calculate_pipe_bend_radius(i_ps: int, radius_fw_channel: float, b_bz_liq: fl
     return radius_pipe_90_deg_bend, radius_pipe_180_deg_bend
 
 
-def set_pumping_powers_as_fractions(
+def pumping_powers_as_fractions(
     f_p_fw_coolant_pump_total_heat: float,
     f_p_blkt_coolant_pump_total_heat: float,
     f_p_shld_coolant_pump_total_heat: float,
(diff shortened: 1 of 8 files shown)


process/models/engineering/ivc_functions.py
```diff
@@ -0,0 +1,307 @@
...
+    b_bz_liq: float
+        radial width of the rectangular cooling channel [m]
+        for long poloidal sections of blanket breeding zone
+
+    """
+    # If primary coolant or secondary coolant (See DCLL)
+    radius_pipe_90_deg_bend = (3 * radius_fw_channel) if (i_ps == 1) else b_bz_liq
+    radius_pipe_180_deg_bend = radius_pipe_90_deg_bend / 2
+
+    return radius_pipe_90_deg_bend, radius_pipe_180_deg_bend
```
timothy-nunn, line 24: Its not immediately clear to me why this function needed to move, was it part of the problem?

process/models/ivc_tools.py
```diff
@@ -0,0 +1,307 @@
...
+
+    """
+    # If primary coolant or secondary coolant (See DCLL)
+    radius_pipe_90_deg_bend = (3 * radius_fw_channel) if (i_ps == 1) else b_bz_liq
+    radius_pipe_180_deg_bend = radius_pipe_90_deg_bend / 2
+
+    return radius_pipe_90_deg_bend, radius_pipe_180_deg_bend
+
+
+def set_pumping_powers_as_fractions(
```
timothy-nunn:
```suggestion
def pumping_powers_as_fractions(
```
I don't like the word `set_` when its not setting anything. We can fix this while we do this move

process/models/ivc_tools.py
```diff
@@ -0,0 +1,307 @@
...
+
+    Parameters
+    ----------
+    rshell :
+
+    rmini :
+
+    rmino :
+
+    zminor :
```
timothy-nunn: Can we fix this docstring to the correct format while we are moving this

process/models/engineering/ivc_functions.py
timothy-nunn, line 1: 1. Can you move this file under `engineering/`
2. Can you rename it to something a bit easier to understand, maybe `blanket_functions.py`

timothy-nunn (changes requested):

process/models/engineering/ivc_functions.py
clmould: It was being used in FW and also some other places too, so made sense to move it to another place along with the other fns here as it didn't seem specific to the blanket module

timothy-nunn (approved):

PR #4244, opened 2026-05-13
Convert Pulse data structure to dataclass

timothy-nunn:


tests/unit/models/test_pulse.py
```diff
@@ -9,19 +9,18 @@
     pf_power_variables,
     pfcoil_variables,
     physics_variables,
-    pulse_variables,
 )
 from process.models.pulse import Pulse
```
clmould: The `Pulse` object should not be used within tests, instead the `pulse` fixture should be used.

clmould (changes requested):

clmould (approved):

PR #4245, opened 2026-05-13
Convert primary pumping data structure to dataclass

timothy-nunn:


process/data_structure/primary_pumping_variables.py
```diff
@@ -1,62 +1,42 @@
...
-    dp_blkt = 3.5e3  # Pa
-    dp_liq = 1.0e7  # Pa
-    p_fw_blkt_coolant_pump_mw = 0.0
-    f_p_fw_blkt_pump = 1.0
+from dataclasses import dataclass
+
+
+@dataclass
+class PrimaryPumpingData:
+    gamma_he: float = 1.667  # Ratio of specific heats  Helium
```
clmould: Do we need the inline comments for all of these when the units are given in the var description already?

clmould (changes requested):

clmould (approved):

PR #4258, opened 2026-05-15
Convert divertor data structure to dataclass

timothy-nunn:


process/models/stellarator/stellarator.py
```diff
@@ -1009,27 +1008,27 @@ def st_fwbs(self, output: bool):
                     self.data.fwbs.p_tf_nuclear_heat_mw = 0.0e0
 
         #  Divertor mass
-        #  N.B. divertor_variables.a_div_surface_total is calculated in stdiv after this point, so will
+        #  N.B. self.data.divertor.a_div_surface_total is calculated in stdiv after this point, so will
```
clmould, line 1011: Is this comment still relevant? I can't find `stdiv` anywhere in the code

clmould (approved):

process/models/stellarator/stellarator.py
timothy-nunn: Neither can I but I think the general sentiment (that a_div_surface_total is 0 on the first iteration) is still true so I would be inclined to keep the comment as is

PR #4286, opened 2026-06-03
Correct output_prefix when bare IN.DAT used

### LINKED ISSUE #4285: Running `IN.DAT` creates two MFILES
Running a file named `IN.DAT` creates two MFILES:
- `.MFILE.DAT`
- `MFILE.DAT`

The former contains the MFILE output, and the latter contains a copy of the input file.

timothy-nunn:
Ensures that:
1. Two MFILES/OUTDAT's can never be created for one input file
2. The MFILE/OUTDAT are correctly named when a `IN.DAT` input file is used


je-cook (approved):
Thanks Tim

PR #4306, opened 2026-06-11
Convert numerics vars to dataclass

clmould:
## Checklist

### DIFF
--- a/process/core/io/plot/solutions.py
+++ b/process/core/io/plot/solutions.py
@@ -21,7 +21,6 @@ import pandas as pd
 import seaborn as sns
 
 from process.core.io.mfile import MFile
-from process.data_structure import numerics
 from process.data_structure.numerics import FiguresOfMerit
 
 # Variables of interest in mfiles and subsequent dataframes
@@ -445,7 +444,6 @@ def _plot_solutions(
     ):
         objf_list = norm_objf_df[NORM_OBJF_NAME].unique()
     else:
-        numerics.init_numerics()
         objf_list = {
             FiguresOfMerit(abs(int(minmax))).description for minmax in diffs_df["minmax"]
         }
--- a/process/core/solver/solver.py
+++ b/process/core/solver/solver.py
@@ -19,7 +19,6 @@ from scipy.optimize import fsolve
 from process.core.exceptions import ProcessValueError
 from process.core.model import DataStructure
 from process.core.solver.evaluators import Evaluators
-from process.data_structure import numerics
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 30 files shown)


process/core/io/plot/solutions.py
```diff
@@ -445,7 +444,6 @@ def _plot_solutions(
     ):
         objf_list = norm_objf_df[NORM_OBJF_NAME].unique()
     else:
-        numerics.init_numerics()
```
timothy-nunn, line 448: This was clearly being done for a reason before. Can you check that plot solutions does not change before and after because of removing this statement.

process/core/solver/solver.py
```diff
@@ -19,19 +19,19 @@
...
-from process.data_structure import numerics
 
 logger = logging.getLogger(__name__)
 
 
 class _Solver(ABC):
     """Base class for different solver implementations."""
 
-    def __init__(self, *, maxcal: int | None = None):
+    def __init__(self, *, data: DataStructure, maxcal: int | None = None):
```
timothy-nunn: If we are passing the data structure in we can remove the `maxcal` argument and just get it from `self.data`

process/core/solver/solver.py
```diff
@@ -359,11 +363,13 @@ def get_solver(data: DataStructure, solver_name: str = "vmcon") -> _Solver:
...
     if solver_name == "vmcon":
-        solver = Vmcon(maxcal=data.globals.maxcal)
+        solver = Vmcon(data=data, maxcal=data.globals.maxcal)
     elif solver_name == "vmcon_bounded":
-        solver = VmconBounded(maxcal=data.globals.maxcal)
+        solver = VmconBounded(data=data, maxcal=data.globals.maxcal)
     elif solver_name == "fsolve":
-        solver = FSolve()
+        solver = FSolve(
+            data=data,
```
timothy-nunn:
```suggestion
            data=data
```

timothy-nunn (changes requested):

process/core/io/plot/solutions.py
clmould: I've found that this else statement used to use `numerics.lablmm`, which has since been replace by `FiguresOfMerit`, which made the use of the `init_numerics` here unnecessary.
 
I have tested this out using the optimum solutions comparison notebook and removing `objf_name` from the MFILEs used (as this is necessary to hit this `else`), and all the plots are the same as the original notebook, so I think it is safe to remove `numerics.init_numerics` from here.

timothy-nunn (approved):

PR #4395, opened 2026-06-29
Particle balance constraint updated

chris-ashe:
This PR add 5 new constraint equation with regard to plasma fuelling and particles loss constituting a way to enforce a plasma particle number balance for all species (except impurities)

image

To do so we now have our plasma fuelling into the vessel as an input parameter whose composition can be changed. The efficiency of this fuelling can be changed also to be indicative of the physical efficiency of the injection.

**Documentation and navigation updates:**
- Added the new plasma fuelling documentation page to the navigation in `mkdocs.yml`, making the new model easily accessible in the documentation site.

**Miscellaneous improvements:**
- Added Avogadro's number to the constants module for use in physical calculations.
- Improved output reporting for constraint errors in `scan.py`.

These changes collectively provide a robust framework for modeling plasma fuelling, tracking individual fuel species, and ensuring physical consistency in fusion plasma simulations.

------------

🎨 Output additions

Expanded and added more values for rates to the fusion reaction summary page:
image

Added a fuelling summary page that shows the contour graphs of the fuelling solutions, along with burnup data:

image

-------------

### Before

image

image

### After

image

image

## Checklist


process/core/io/plot/summary.py
```diff
@@ -11771,15 +11779,15 @@ def plot_fusion_rate_profiles(axis: plt.Axes, fig, mfile: MFile, scan: int):
...
-        f"Rate density, plasma: {mfile.get('fusden_plasma_alpha', scan=scan):.4e} particles/m3/sec\n\n"
-        f"Total power density: {mfile.get('pden_alpha_total_mw', scan=scan):.4e} MW/m3\n"
-        f"Plasma power density: {mfile.get('pden_plasma_alpha_mw', scan=scan):.4e} MW/m3\n\n"
-        f"Power per unit volume transferred to electrons: {mfile.get('f_pden_alpha_electron_mw', scan=scan):.4e} MW/m3\n"
-        f"Power per unit volume transferred to ions: {mfile.get('f_pden_alpha_ions_mw', scan=scan):.4e} MW/m3\n\n"
+        f"Total power: {mfile.get('p_alpha_total_mw', scan=scan):.4f} MW\n"
+        f"Plasma power: {mfile.get('p_plasma_alpha_mw', scan=scan):.4f} MW\n"
+        f"Beam power: {mfile.get('p_beam_alpha_mw', scan=scan):.4f} MW\n\n"
+        f"Rate density total: {mfile.get('fusden_alpha_total', scan=scan):.4e} particles/m$^3$/sec\n"
+        f"Rate density, plasma: {mfile.get('fusden_plasma_alpha', scan=scan):.4e} particles/m$^3$/sec\n\n"
```
j-a-foster: Why are these /sec and not /s?

process/core/io/plot/summary.py
```diff
@@ -15300,6 +15309,311 @@ def plot_cs_radial_stress_profile(
...
+    axis.contour(
+        fuelling_range,
+        recycling_range,
+        tritium_flow,
+        levels=[0],
+        colors="black",
+        linewidths=2,
+    )
+
+    # Plot star for mfile values
```
j-a-foster, line 15927: Is it worth adding a note or legend that explains what the star means in the PDF?

j-a-foster (changes requested):
Just a couple of notes, otherwise happy with the changes.

process/models/physics/fuelling.py
```diff
@@ -0,0 +1,236 @@
+import logging
+
+from process.core import constants
+from process.core.model import Model
+
+logger = logging.getLogger(__name__)
+
+
+class PlasmaFuelling(Model):
+    """Class to hold plasma fuelling calculations for plasma processing."""
```
jonmaddock: Plasma processing? Could you be a bit more descriptive here?

process/models/physics/fuelling.py
```diff
@@ -0,0 +1,236 @@
...
+            Ratio of alpha particle confinement time to energy confinement time (dimensionless).
+        nd_plasma_alphas_vol_avg : float
+            Volume-averaged density of alpha particles in the plasma (particles/m^3).
+        vol_plasma : float
+            Plasma volume (m^3).
+
+        Returns
+        -------
+        float
+            Alpha particle flow rate in the plasma exhaust (particles/s).
```
jonmaddock: Isn't this the wrong way around? It's the flow rate into the plasma (for a positive returned value), not the exhaust?

process/models/physics/fuelling.py
```diff
@@ -0,0 +1,236 @@
...
+    @staticmethod
+    def calculate_plasma_alphas_flow_rate(
+        fusrat_dt_total: float,
+        fusrat_plasma_dhe3: float,
+        t_energy_confinement: float,
+        f_t_alpha_energy_confinement: float,
+        nd_plasma_alphas_vol_avg: float,
+        vol_plasma: float,
+    ) -> float:
+        """Calculate the alpha particle flow rate in the plasma exhaust.
```
jonmaddock: Shouldn't these be in the exhaust model?

process/models/physics/physics.py
```diff
@@ -633,6 +633,22 @@ def run(self):
...
+            + self.data.physics.fusrat_plasma_dd_triton
+        )
+
+        self.data.physics.fusrat_neutron_production_total = (
+            self.data.physics.fusrat_plasma_dd_helion + self.data.physics.fusrat_dt_total
+        )
+
+        self.data.physics.fusrat_dt_total = (
+            self.data.physics.p_dt_total_mw * 1e6 / (constants.D_T_ENERGY)
+        )
```
jonmaddock, line 694: Are all of these not already calculated elsewhere?

process/models/physics/fuelling.py
```diff
@@ -19,6 +19,99 @@ def output(self):
...
+        f_molflow_plasma_fuelling_tritium : float
+            Fraction of tritium in the plasma fuelling.
+
+        Returns
+        -------
+        Tritium burnup fraction (dimensionless).
+
+        Notes
+        -----
+        The fusion rate is multiplied by two to convert from nucleus pairs to particles,
```
jonmaddock, line 126: Not multiplied by 2 here.

process/main.py
```diff
@@ -771,6 +774,7 @@ def models(self) -> tuple[Model, ...]:
             self.plasma_fields,
             self.sauter_bootstrap_current,
             self.plasma_transition,
+            self.self.plasma_fuelling,
```
jonmaddock: `self.self`

process/core/solver/constraints.py
```diff
@@ -1762,6 +1762,153 @@ def constraint_equation_92(constraint_registration, data):
...
+
+    This constraint ensures that the tritium particle balance is maintained in the
+    plasma.
+
+    The numerator represents the total tritium sources, including fuelling and fusion
+    reactions, while the denominator represents the total tritium sinks, which are
+    related to the plasma volume, fuel ion density, and confinement time.
+
+    The equation ensures that the rate of tritium production equals the rate of tritium
+    loss, maintaining a steady-state condition for tritium in the plasma.
```
jonmaddock, line 1918: Great description.

process/core/solver/constraints.py
```diff
@@ -1762,6 +1762,153 @@ def constraint_equation_92(constraint_registration, data):
...
+def constraint_equation_93(constraint_registration, data):
+    """
+    Tritium particle balance consistency equation.
+
+    This constraint ensures that the tritium particle balance is maintained in the
+    plasma.
+
+    The numerator represents the total tritium sources, including fuelling and fusion
+    reactions, while the denominator represents the total tritium sinks, which are
+    related to the plasma volume, fuel ion density, and confinement time.
```
jonmaddock, line 1915: Net tritium production over confinement loss rate?

process/core/solver/constraints.py
```diff
@@ -1762,6 +1762,153 @@ def constraint_equation_92(constraint_registration, data):
...
+        + data.physics.fusrat_plasma_dd_triton
+        - data.physics.fusrat_dt_total
+    )
+    denominator = (
+        data.physics.nd_plasma_fuel_ions_vol_avg
+        * data.physics.vol_plasma
+        * data.physics.f_plasma_fuel_tritium
+    ) / (
+        data.physics.t_energy_confinement
+        / (1 - data.physics.f_plasma_particles_lcfs_recycled)
```
jonmaddock: Ref for this "effective" confinement time due to recycling?

process/core/solver/constraints.py
```diff
@@ -1762,6 +1762,153 @@ def constraint_equation_92(constraint_registration, data):
...
+        data.physics.eta_plasma_fuelling
+        * data.physics.molflow_plasma_fuelling_vv_injected
+        * data.physics.f_molflow_plasma_fuelling_helium3
+    ) + data.physics.fusrat_plasma_dhe3
+    denominator = (
+        data.physics.nd_plasma_fuel_ions_vol_avg
+        * data.physics.vol_plasma
+        * data.physics.f_plasma_fuel_helium3
+    ) / (
+        data.physics.t_energy_confinement
```
jonmaddock: Does He3 follow the energy confinement time rather than the tau_alpha / tau_E = 5 relation?

documentation/source/physics-models/plasma_fuelling.md
```diff
@@ -0,0 +1,209 @@
...
+
+
+
+### METIS Alpha Confinement
+
+$$
+\tau_{\alpha} = f_{\alpha}\tau_{\text{E}}\frac{R}{1-R}\tau_{\text{ne}}
+$$
+
+This is the model currently in METIS[^2] and is found in [^3]
```
jonmaddock: But this isn't used in Process?

documentation/source/physics-models/plasma_fuelling.md
```diff
@@ -0,0 +1,209 @@
...
+
+
+--------------
+
+### Tritium Flow Rate | `calculate_plasma_tritium_flow_rate()`
```
jonmaddock, line 61: Not sure about "flow rate": it's the density derivative in the plasma.

documentation/source/physics-models/plasma_fuelling.md
```diff
@@ -0,0 +1,209 @@
...
+
+This constraint can be activated by stating `icc = 97` in the input file.
+
+This ensures that all 3 injected fuelling fractions sum up to 1:
+
+$$
+f_{\text{fuelling,D}} + f_{\text{fuelling,T}} + f_{\text{fuelling,3He}} = 1.0
+$$
+
+**It is recommended to have this constraint on as it is a plasma consistency model**
```
jonmaddock, line 235: I think you need to be more explicit about the system of equations (i.e. all of the above constraints) and the solution parameters (i.e. optimisation parameters) used to solve them. What's should the user do to enforce all of these constraints in their optimisation problem?

process/core/io/plot/summary.py
```diff
@@ -15457,6 +15771,21 @@ def _add_page(name: str | None = None):
...
+    )
+    plot_helium3_flow_contour(
+        axis=pages["fuelling"].add_subplot(233), mfile=m_file, scan=scan
+    )
+    plot_alpha_flow_contour(
+        axis=pages["fuelling"].add_subplot(223), mfile=m_file, scan=scan
+    )
+
+    plot_fuelling_info(pages["fuelling"], m_file, scan)
+
```
jonmaddock, line 16366: Can I see these plots in the PR please?

jonmaddock (changes requested):
- [ ] 1. Please sort out the PR description: I think it's too verbose and misses/buries the key point which is that you're introducing constraints to enforce density equilibrium of individual ion species. Why are you making this PR?
- [x] 2. Have you removed the original `molflow_plasma_fuelling_required` as an output? I might have missed that
- [ ] 3. I counted 6 new optimisation parameters and 5 new constraints: as per our conversation, can you describe how this might work in solution mode, i.e. when we require a determined system?
- [ ] 4. Some plots to demonstrate these changes would be useful: for example how the constraints are accommodated with increasing te or ne, for example. How does this change the current large tokamak solution?
- [ ] 5. How does this compare to the existing burnup calculation? Has it been removed?
- [ ] 6. I'd like this PR to include what equations and parameters should be included in optimisation and solution scenarios. If the solution system is under-determined, how useful is it?
- [x] 7. I'm not sure about the fuelling composition constraint and total fuelling rate. Would individual species rates reduce the dimensionality?
- [ ] 8. Should the recycling fraction and fuelling efficiency be optimisation parameters? (I realise we've discussed this, but I think it should be made clear why these can be used to solve the constraints).

The created docs were excellent. Thanks for the fixed-up commits, this was much easier to review.

chris-ashe:
@timothy-nunn If I try and use the functions in `fuelling.py` instead of re-writing the equations explicitly in `constraint.py` I get an `inf` error on the first iteration, would you be able to look at this as we had problems before about the constraint value

documentation/source/physics-models/plasma_fuelling.md
chris-ashe: I left it as flow rate as the functions returns the net particle flow in the plasma, so not necessarily the density

process/core/solver/constraints.py
chris-ashe: No, the descriptions say the are just related/proportional to. In reality the loss term as seen in the docs is the total particle number over the effective particle confinement time

process/core/solver/constraints.py
chris-ashe: The docs reference:

- J. F. Artaud et al., “Metis: a fast integrated tokamak modelling tool for scenario design,” Nuclear Fusion, vol. 58, no. 10, pp. 105001–105001, Aug. 2018, doi: https://doi.org/10.1088/1741-4326/aad5b1.

- D. Reiter, H. Kever, G. H. Wolf, M. Baelmans, R. Behrisch, and R. Schneider, “Helium removal from tokamaks,” Plasma Physics and Controlled Fusion, vol. 33, no. 13, pp. 1579–1600, Nov. 1991, doi: https://doi.org/10.1088/0741-3335/33/13/008.

With the former being the calculation used in METIS which uses the R recycling coefficient also to define the effective confinement time.

Waiting for Tim to check a solver `inf` error so that the equations can be used explicitly here with the references attached instead of writing them out again here

process/core/solver/constraints.py
chris-ashe: This is a point we should discuss, I dont see why we wouldn't treat it the same as 4He

process/models/physics/physics.py
chris-ashe: No, I had to add them. We have values for the powers but all of the actual rates were done in situ before. There is changes to the fusion power summary page which I will add to the PR overview also

process/models/physics/fuelling.py
chris-ashe: The comment has been ammended, since it works the exact same as the other fuel models and returns the source rate, I think it is best to leave it with the group.

process/core/io/plot/summary.py
chris-ashe: Will add

documentation/source/physics-models/plasma_fuelling.md
```diff
@@ -0,0 +1,209 @@
...
+\frac{dn_{\alpha}}{dt} = \Gamma_{\text{D+3He}} + \Gamma_{\text{D+T}} - \frac{N_{\alpha}}{\tau_{\alpha}^*}
+$$
+
+In a steady state equilibrium all 4 of these equations should balance, therefore:
+
+$$
+\frac{dn_{\text{D}}}{dt} = \frac{dn_{\text{T}}}{dt} = \frac{dn_{\text{3He}}}{dt} = \frac{dn_{\alpha}}{dt} = 0
+$$
+
+Here $\eta_{\text{fuelling}}$ is the fuelling efficiecny which represents the method of injecting fuel into the plasma. Gas puffing on the low field side is probably around 0.01-0.1, supersonic gas is 0.1 and 0.2 and using pellets can get you close to unity with 0.5-0.9. $\Gamma_{\text{fuelling}}$ is the fuel injection rate into the vacuum vessel, so $\eta_{\text{fuelling}} \Gamma_{\text{fuelling}}$ together presents the fraction of injected fuel that actually makes it into the plasma core to fuse.
```
grmtrkngtn:
```suggestion
Here, $\eta_{\text{fuelling}}$ is the fuelling efficiency, which quantifies how effectively fuel injected into the vacuum vessel reaches the plasma core. The value of $\eta_{\text{fuelling}}$ depends on the fuelling method. Typical values are around 0.01--0.1 for low-field-side gas puffing, 0.1--0.2 for supersonic gas injection, and 0.5--0.9 for pellet injection, which can approach unity under favourable conditions. $\Gamma_{\text{fuelling}}$ is the fuel injection rate into the vacuum vessel. The product $\eta_{\text{fuelling}} \Gamma_{\text{fuelling}}$ therefore represents the effective fuelling rate, i.e. the fraction of the injected fuel that successfully penetrates into the plasma and becomes available for fusion reactions.
```

grmtrkngtn: Suggestion to tidy up this section a bit.

documentation/source/physics-models/plasma_fuelling.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Fuel Burnup Fration
+
+In a steady state tokamak, the burnup fraction ($f_b$) is explicitly defined by the following rate equation:
+
+$$
+f_b = \frac{\Gamma_{\text{fusion}}}{\Gamma_{\text{fuel}}}
+$$
+
+where $\Gamma_{\text{fusion}}$ is the fusion rate in reactions per second $[\text{s}^{-1}]$, and $\Gamma_{\text{fuel}}$ is our fuelling rate into the vessel like is shown above also in $[\text{s}^{-1}]$.
```
grmtrkngtn:
```suggestion
where $\Gamma_{\text{fusion}}$ is the fusion reaction rate per second $[\text{s}^{-1}]$, and $\Gamma_{\text{fuel}}$ is the fuel injection rate into the vacuum vessel in particles per second $[\text{s}^{-1}]$.
```

process/models/physics/fuelling.py
```diff
@@ -0,0 +1,572 @@
...
+        as the fuelling rate is in particles/s.
+
+        """
+        return fusrat_dt_total / (
+            molflow_plasma_fuelling_vv_injected * f_molflow_plasma_fuelling_tritium
+        )
+
+    @staticmethod
+    def calculate_deuterium_burnup_fraction(
```
grmtrkngtn, line 136: There's an issue for beam target fusion tokamaks here, like VNS. 

In a D beam case, the reaction is D_beam + T_thermal. Therefore thermal tritium is consumed but not thermal deuterium.

In this case, we cannot use the total DT rate, as it would overestimate D_thermal consumption, and therefore the required D fuelling.

Could we separate thermal DT from beam-target DT and create a beam-target sink according to the beam isotope fraction?

```thermal_d_dt_consumption = (fusrat_dt_thermal + f_beam_tritium * fusrat_dt_beam)```

Something like the above.

File: process/models/physics/fuelling.py [hunk 20]
```diff
@@ -0,0 +1,572 @@
...
+        Notes
+        -----
+        The fusion rate is multiplied by two to convert from nucleus pairs to particles,
+        as the fuelling rate is in particles/s.
+
+        """
+        return 2 * fusrat_total / molflow_plasma_fuelling_vv_injected
+
+    @staticmethod
+    def calculate_tritium_burnup_fraction(
```
grmtrkngtn, line 101: Also here we need to make a modification for the beam-fusion/VNS case.

For a D beam, both thermal–thermal DT and beam-target DT consume thermal tritium. For a T beam, the beam-target reaction consumes thermal deuterium instead.

process/models/physics/fuelling.py
```diff
@@ -0,0 +1,572 @@
...
+        while a negative value indicates a net loss of helium-3 from the plasma.
+
+        """
+        return (
+            (
+                f_molflow_plasma_fuelling_helium3
+                * eta_plasma_fuelling
+                * molflow_plasma_fuelling_vv_injected
+            )
+            + fusrat_plasma_dhe3
```
grmtrkngtn: Should this be a subtraction and not an addition?  We're fuelling with He3 and consuming with fusion.

process/models/physics/fuelling.py
```diff
@@ -0,0 +1,572 @@
...
+            fusrat_plasma_dhe3=self.data.physics.fusrat_plasma_dhe3,
+        )
+
+    def output(self):
+        """This model doesn't output to the output file, but it does generate contour
+        plots of plasma fuel flow rates vs recycling and fuelling efficiency.
+        """
+
+    @staticmethod
+    def calculate_fuel_burnup_fraction(
```
grmtrkngtn, line 61: This requires thought for the beam-fusion reaction case. This currently assumes that the fusion reactions only come from plasma-fuelling, but in the case where we have beam-fusion reactions  only one (or some fraction) of the plasma fuelling ions are consumed for each species.

process/core/solver/constraints.py
```diff
@@ -1794,6 +1795,150 @@ def constraint_equation_92(constraint_registration, data):
...
+    related to the plasma volume, fuel ion density, and confinement time.
+
+    The equation ensures that the rate of helium-3 production equals the rate of helium-3
+    loss, maintaining a steady-state condition for helium-3 in the plasma.
+    """
+    numerator = (
+        data.physics.eta_plasma_fuelling
+        * data.physics.molflow_plasma_fuelling_vv_injected
+        * data.physics.f_molflow_plasma_fuelling_helium3
+    ) + data.physics.fusrat_plasma_dhe3
```
grmtrkngtn: Similar to comment in fuelling - is dhe3 a source? Shuldn't it be subtracted?

process/core/solver/constraints.py
grmtrkngtn, line 1: Constraints:

The way the constraints are written is a bit inconsistent. Some of them use helper functions from `PlasmaFuelling`, but the D and He3 constraints reconstruction the reaction and confiment in the constraint.

I think ideally we wouldn't perform these calculations in the constraint equations, and they are calculated somewhere else in the code. This would help avoid inconsistencies, particular in beam-fusion case where we have different sources/sinks or branching calculations to consider.

Could the constraints be re-written to use the balance terms already calculated by PlasmaFuelling rather than independently re-calculating them?

grmtrkngtn (changes requested):
I have a few questions about how this will handle a beam-fusion reactor, as this introduces some additional complications.

First, the thermal fuel mix is no longer necessarily 50/50, and the beam introduces both an additional fuel source and an additional fusion sink. I have left some comments in `PlasmaFuelling` on how the beam-target contribution could be accounted for in the species balances.

My understanding is that the existing PROCESS composition logic starts from electron density, calculates the total fuel-ion density, and then derives the individual D, T and He3 densities from the prescribed fuel fractions. If that is still the case, the composition routine may overwrite or constrain the same species densities that these new particle-balance constraints are intended to solve.

Have you also updated the density closure so that the absolute D and T densities can vary independently, with electron density and the fuel fractions then derived from charge neutrality? Otherwise, I am not sure that the D and T balances can act as independent solution constraints.

process/core/solver/constraints.py
chris-ashe: I knew I forgot to update something....

Yes, all the constraints should call the source and sink functions like is done for the Tritium at the moment. Me and Tim were experimenting with solver errors when I wrote the tritium one so forgot to do the rest. Will add those in

process/core/solver/constraints.py
chris-ashe: Yes, this is muddled around, have fixed now

process/models/physics/fuelling.py
chris-ashe: This will require a bit more work and most likely other constraints for the rate of fast D and T thermalisation. A bit like what will be added for the fast alphas. At the moment the total amount of D and T assumptions is just measured

process/models/physics/fuelling.py
chris-ashe: Have added the fuelling components for the tritium beam

process/models/physics/fusion_reactions.py
grmtrkngtn, line 1040: Should we add these to output? Might be useful/interesting.

process/models/physics/physics.py
grmtrkngtn: fusrat_neutron_production_total should be calculated after fusrat_dt_total since it uses it.

process/models/physics/physics.py
```diff
@@ -1838,6 +1789,77 @@ def outplas(self):
...
+            "OP ",
+        )
+        po.ovarre(
+            self.outfile,
+            "Neutron production rate: total (particles/sec)",
+            "(fusrat_neutron_production_total)",
+            self.data.physics.fusrat_neutron_production_total,
+            "OP ",
+        )
+        po.ovarre(
```
grmtrkngtn: This is printed already at line 1828 - remove this instance (or the other)

process/models/physics/physics.py
```diff
@@ -1838,6 +1789,77 @@ def outplas(self):
...
+            self.outfile,
+            "D-3He Fusion rate: total (reactions/sec)",
+            "(fusrat_plasma_dhe3)",
+            self.data.physics.fusrat_plasma_dhe3,
+            "OP ",
+        )
+        po.ovarre(
```
grmtrkngtn: Output is duplicated, first instance is just above.

process/models/physics/physics.py
```diff
@@ -1838,6 +1789,77 @@ def outplas(self):
...
+            self.outfile,
+            "Neutron production rate: total (particles/sec)",
+            "(fusrat_neutron_production_total)",
+            self.data.physics.fusrat_neutron_production_total,
+            "OP ",
+        )
+        po.ovarre(
```
grmtrkngtn: This is also output twice.

process/models/physics/fusion_reactions.py
chris-ashe: We do have `nd_plasma_fuel_ions_vol_avg` which is the total fuel ion mix but not the individual species. May put a PR up to add this in a extra

process/models/physics/physics.py
chris-ashe: Have now moved to the correct order

documentation/source/physics-models/plasma_fuelling.md
jonmaddock: Snippet of constraints and opt params required to enable this please.

jonmaddock (changes requested):
Questions not yet answered, changes made unclear due to force-push.

timothy-nunn:
@chris-ashe for a PR of this size that takes a long time to review you can try using `git merge` rather than `git rebase` to resolve conflicts. This _should_ allow GitHub to keep track of the changes better than a force push and might help Jon et al see what changes have been made when you push!

chris-ashe:
> * [ ]  Please sort out the PR description: I think it's too verbose and misses/buries the key point which is that you're introducing constraints to enforce density equilibrium of individual ion species. Why are you making this PR?
> * [x]  Have you removed the original `molflow_plasma_fuelling_required` as an output? I might have missed that
> * [ ]  I counted 6 new optimisation parameters and 5 new constraints: as per our conversation, can you describe how this might work in solution mode, i.e. when we require a determined system?
> * [ ]  Some plots to demonstrate these changes would be useful: for example how the constraints are accommodated with increasing te or ne, for example. How does this change the current large tokamak solution?
> * [ ]  How does this compare to the existing burnup calculation? Has it been removed?
> * The created docs were excellent
> * [ ]  I'd like this PR to include what equations and parameters should be included in optimisation and solution scenarios. If the solution system is under-determined, how useful is it?
> * [x]  I'm not sure about the fuelling composition constraint and total fuelling rate. Would individual species rates reduce the dimensionality?
> * [ ]  Should the recycling fraction and fuelling efficiency be optimisation parameters? (I realise we've discussed this, but I think it should be made clear why these can be used to solve the constraints).
> 
> Thanks for the fixed-up commits, this was much easier to review.

- For setting up the problem in the `_eval` file we now have 3 particle balance constraints on and the fuelling composition constraint. There is now 4 new added iteration variables in the file. The fuelling efficiency rate is now prescribed.

- In terms of the solution for the optimised large tokamaks solution it looks like none of the original values have changed but the fuelling constraints have accommodated to get the same answer

- For the burnup in the optimised large tokamak solution the burnup is 17.94% while with the new model the calculates it directly we get a total fuel burnup of 23.4%. We can also calculate the individual burnup of each species.

- I have added some more detail to the docs about setting up the optimisation and solution scenarios

jonmaddock:
I've numbered my questions now to aid the discussion.

1. Not done
3. Not sure if this has been answered in the docs (your 4th bullet) as I can't see the changes, but can you answer this here?
4. I think the "before and after" plots in the description partly answer this. Can you make clear what has changed in the "after" case (i.e. which constraints and optimisation parameters were added)? Given that the solution doesn't really change much, can you describe what is changing in the description (i.e. the fuelling rate, as in your 2nd bullet). Can you explain when these changes *would* have an effect on the solution (e.g. a machine with more asymmetric fuelling), and include this is the description?
5. This has been answered in your 3rd bullet, but has the old burnup calculation been removed? That's a fair increase! Is it right/agrees with known values?
6. Not done
8. Maybe in docs, but should be made clear here. Not done

I don't understand your 1st bullet: it's not clear to me what the system of equations now is. I think, as we discussed, it might be worth making a separate eval file that includes these modifications, so we have a "general" problem definition as we currently have and a separate "individual species" problem.

chris-ashe:
> I've numbered my questions now to aid the discussion.
> 
> 1. Not done
> 2. Not sure if this has been answered in the docs (your 4th bullet) as I can't see the changes, but can you answer this here?
> 3. I think the "before and after" plots in the description partly answer this. Can you make clear what has changed in the "after" case (i.e. which constraints and optimisation parameters were added)? Given that the solution doesn't really change much, can you describe what is changing in the description (i.e. the fuelling rate, as in your 2nd bullet). Can you explain when these changes _would_ have an effect on the solution (e.g. a machine with more asymmetric fuelling), and include this is the description?
> 4. This has been answered in your 3rd bullet, but has the old burnup calculation been removed? That's a fair increase! Is it right/agrees with known values?
> 5. Not done
> 6. Maybe in docs, but should be made clear here. Not done
> 
> I don't understand your 1st bullet: it's not clear to me what the system of equations now is. I think, as we discussed, it might be worth making a separate eval file that includes these modifications, so we have a "general" problem definition as we currently have and a separate "individual species" problem.

2. Yes, `molflow_plasma_fuelling_required` is now gone and replaced with `molflow_plasma_fuelling_vv_injected`, which is now our actual control optimisation parameter. So we give it as an input instead of letting it be calculated.
3. In the large tokamak optimisation case the final optimal solution has not changed. For the optimisation cases the number of equality constraints active has increased by 4. This is for the Tritium, Deuterium, Thermal Alpha, and fuelling compisition consistency. This is the same for the `eval` regression files.  For the optimisation :
```

n_equality_constraints = 7

* Tritium particle balance
icc = 93

* Deuterium particle balance
icc = 94

* Alpha particle balance
icc = 96

* Fuelling composition consistency
icc = 97


----------------


* Particle recycling fraction
ixc = 178
f_plasma_particles_lcfs_recycled = 0.9

* Plasma fuelling efficiecy
ixc = 179
eta_plasma_fuelling = 0.7

* Injected VV fuelling rate
ixc = 180
molflow_plasma_fuelling_vv_injected = 5e21
boundl(180) = 1e20

* Deuterium fuelling fraction
ixc = 181
f_molflow_plasma_fuelling_deuterium = 0.5
boundl(181) = 0.4

* Tritium fuelling fraction
ixc = 182
f_molflow_plasma_fuelling_tritium = 0.5
boundl(182) = 0.4
```

We have the fuelling mixture, fuelling rate, fuelling efficiency and recycling coefficient as iteration variables.
For the `eval` case:
```

n_equality_constraints = 6

* Tritium particle balance
icc = 93
* Deuterium particle balance
icc = 94
* Alpha particle balance
icc = 96
* Fuelling composition consistency
icc = 97

-----------


* Particles recycled at LCFS
ixc = 178
f_plasma_particles_lcfs_recycled = 0.89

* Injected VV fuelling rate
ixc = 180
molflow_plasma_fuelling_vv_injected = 4.9992093e21

* Deuterium fuelling fraction
ixc = 181
f_molflow_plasma_fuelling_deuterium = 0.50092526

* Tritium fuelling fraction
ixc = 182
f_molflow_plasma_fuelling_tritium = 0.49907474
```
We have the same except we have the fuelling efficiency input so that our new number of iterations vars matches our increase increase in constraints (+4).

4. Yes the old burnup calc is now gone as we now calculate it implicitly based on its actual definition. Our old function used a guess via using the thermal alpha density, total ion density and the fusion rate with `tauratio` to make a loss term. This is not truly correct as you need to know the rate of creation vs the rate of injection. Our constraints now ensure that we know the true required rate of injection for all species. Therefore we can now calculate the equation implicitly as we already have the fusion rate for each species.  The paper by Jackson "AN ANALYTIC EXPRESSION FOR THE TRITIUM
BURNUP FRACTION IN BURNING-PLASMA DEVICES" tries to use the fuelling efficiency and recycling fraction but makes the use of parametrised profile values (that only work for parabolic) and other assumptions. In its verbose form for equation (1) he creates a reduced form of the equation he originally states which we actually do fully:
image.

In terms of the value for the burnup this can be heavily swayed by the recycling coefficient $R$ which really determines the system throughput. We could try and match the same value of $R$ to get the same burnup as before and it will tell us what our assumed recycling fractions would need to have been for our previous solutions

I will update the PR text to be more concise and give a problem setup for the input files @jonmaddock

jonmaddock:
Ok, thanks @chris-ashe . So,

1. Done
2. Done
3. This is much clearer, thanks, but can you be more explicit? Include the full list of constraints as well as optimisation parameters for the optimisation and "eval" (solution) cases? Can you comment on why the fuelling efficiency is allowed to vary in the optimisation case but not in the solution case, and what it means (pellet to puff ratio, I think you said)?
4. Thanks for including the before and after plots, but I can't see any change in solution. Perhaps you could comment on what the difference is (however small), and under what circumstances these additional constraints and optimisation parameters would give a more different solution to the status quo?
5. (Burnup): Ok, does this mean that existing fuel ion equilibrium-only constraint doesn't work? Do we always need all of these new constraints on to make this new burnup calculation correct?
6. Is this done? 
7. Done
8. Not done

Yours pedantically,

Jon

chris-ashe:
> Ok, thanks @chris-ashe . So,
> 
> 1. Done
> 2. Done
> 3. This is much clearer, thanks, but can you be more explicit? Include the full list of constraints as well as optimisation parameters for the optimisation and "eval" (solution) cases? Can you comment on why the fuelling efficiency is allowed to vary in the optimisation case but not in the solution case, and what it means (pellet to puff ratio, I think you said)?
> 4. Thanks for including the before and after plots, but I can't see any change in solution. Perhaps you could comment on what the difference is (however small), and under what circumstances these additional constraints and optimisation parameters would give a more different solution to the status quo?
> 5. (Burnup): Ok, does this mean that existing fuel ion equilibrium-only constraint doesn't work? Do we always need all of these new constraints on to make this new burnup calculation correct?
> 6. Is this done?
> 7. Done
> 8. Not done
> 
> Yours pedantically,
> 
> Jon

3. I have ammended my previous comment to now show the constraints and total constraint number. The `_eval` case requires equal constraints and iterations variables, while in `_optimisation` case does not. The fuelling efficiency is allowed to change just to ease finding a solution. The problem is very open at the moment as their is no tritium plant or proper vacuum constraints applied to the fuelling inlet

4. For the `_optimisation` case their is no change to the solution for `large_tokamak` at the moment as the problem is open ended, meaning that the tritium plant and vacuum pumps arent putting a min/max on our fuelling inlet and outlet. When these are added we will probably see a change in fusion power.

5. Not sure what you mean by "mean that existing fuel ion equilibrium-only constraint doesn't work". For the burnup, since we know have individual species balances we now also have individual fuel species burnups (see output) these culminate together to make the total fuel burnup fraction.  You only need the constraints on for the fuel species present in the plasma as it will enforce in=out. In most cases the 3He will not be needed as everyone just runs D-T. All current equations look for an exact analytical equation for the burnup which dosent exist due to the burnup problem requiring more than 1 ODE to solve for the full plasma case.

6. Please inspect the bottom of `fuelling.md` which contains a problem set up guide. 

8. For the recycling fraction ($R$) and the fuelling efficiency the former is a higher fidelity unknown and the latter is a machine configuration facsimile. The recycling fraction normally has to be done by higher fidelity plasma-neutral modeling  codes that couple the FW, Divertor and plasma interaction, we cannot do this here. The user will have to assume what their value is for their solution. Treat this the same as making the alpha to energy confinement ratio being equal to 5, its a prior higher order assumption. The fuelling efficiency will be based on what fuelling technology is used. This again is a 3D higher order number. What this means is we can now say "we require a fuelling efficiency > than ... etc"""

jonmaddock:
3. Great. But
>  The fuelling efficiency is allowed to change just to ease finding a solution
Isn't a great reason. What does the fuelling efficiency actually mean? Is this something that can actually be changed at will, or is it an unknown? Otherwise this is a similar argument to having H-factor as an optimisation parameter.

4. Great, thanks. Done.
5. What I mean is does the existing fuel ion equilibrium constraint work by itself as it used to, using the original burnup fraction calculation? We may not necessarily want to enforce individual species equilibrium when trying to find an initial solution, as this will quite likely make it harder to converge. It also removes uncertain parameters (e.g. recycling fraction, fuelling efficiency) that a user may not know accurately/want to specify at an early stage of a design.
6. Great, done.
8. (Should the recycling fraction and fuelling efficiency be optimisation parameters?). I think this means that R and fuelling efficiency should be fixed, not used as parameters to solve these additional equations. There is a value of R, we just don't know it (form HiFi codes): that doesn't mean we get to pick its value to make our equations work and optimise.

chris-ashe:
> 3. Great. But
> 
> > The fuelling efficiency is allowed to change just to ease finding a solution
> > Isn't a great reason. What does the fuelling efficiency actually mean? Is this something that can actually be changed at will, or is it an unknown? Otherwise this is a similar argument to having H-factor as an optimisation parameter.
> 
> 4. Great, thanks. Done.
> 5. What I mean is does the existing fuel ion equilibrium constraint work by itself as it used to, using the original burnup fraction calculation? We may not necessarily want to enforce individual species equilibrium when trying to find an initial solution, as this will quite likely make it harder to converge. It also removes uncertain parameters (e.g. recycling fraction, fuelling efficiency) that a user may not know accurately/want to specify at an early stage of a design.
> 6. Great, done.
> 7. (Should the recycling fraction and fuelling efficiency be optimisation parameters?). I think this means that R and fuelling efficiency should be fixed, not used as parameters to solve these additional equations. There is a value of R, we just don't know it (form HiFi codes): that doesn't mean we get to pick its value to make our equations work and optimise.

3. The fuelling efficiency is necessary component of the equation as in reality no plasma fuelling system is 100% perfect. Gas puffing is around 10% efficient and pellets are around up to 70%. For modelling this means we will always have a waste amount of fuel that will not penetrate the plasma and will need to be recycled in the divertor. This gives us realistic demand requirements for the vacuum pumps when they are properly integrated. In terms of it being an optimisation, allowing it to change can be seen as a facsimile for changing your fuelling method mix on the machine. Fuelling efficiency itself is too complex for us to model. The second benefit of having it is that the model can say we need a minimum efficiency to find solutions for an equilibrium, as a lower efficiency means more demand on the vacuum pumps which have their own limits. So we can say something like: "We need a fuelling efficiency of at least X in order to work with pump type Y. We currently have turbomolecular and cryopumps in our vacuum files which all have different pumping speeds for different species.

5. Which constraint do you mean exactly? 

7. Again, this is similar to our alpha confinement ratio of 5. Realistically the relative fuelling efficiency can be changed in the control room via mass flow into the different systems, eg gas puffing and in pellets. The relative efficiency can also depend on the fuelling location control such as where on the plasma surface to puff or where in the core to aim the pellet. In terms of the recycling coefficient, most machines work in high recycling ($R = 0.9 - 0.99$) where the FW and divertor tiles are already full saturated so an escaped particle strikes the FW or divertor and is neutralised and returns to the plasma rather than being absorbed or pumped away. Advanced pumping with cryopumps can get $R$ down to around 0.8.

PR #4396, opened 2026-06-29
Correct hstar being stored on data structure

timothy-nunn:
I think `hstar` is being incorrectly set on the data structure in `calculate_confinement_time`. `calculate_confinement_time` is also called when creating the H-factor table for the OUT.DAT. I believe that the `hstar` from this is what was being written to the MFILE, not the `hstar` from the selected confinement scaling.

### DIFF
--- a/process/models/physics/confinement_time.py
+++ b/process/models/physics/confinement_time.py
@@ -37,6 +37,8 @@ class ConfinementTimeData:
     """Plasma energy confinement time in seconds."""
     p_plasma_loss_mw: float
     """Total plasma loss power in MW."""
+    hstar: float
+    """Non-radiation corrected H factor on energy confinement times"""
 
 
 class PlasmaConfinementTime(Model):
@@ -936,8 +938,11 @@ class PlasmaConfinementTime(Model):
 
         # Calculate H* non-radiation corrected H factor
         # Note: we will assume the IPB-98y2 scaling.
-        if self.data.physics.i_rad_loss == ConfinementRadiationLossModel.CORE_ONLY:
-            self.data.physics.hstar = (
+        if (
+            ConfinementRadiationLossModel(self.data.physics.i_rad_loss)
+            == ConfinementRadiationLossModel.CORE_ONLY
+        ):
+            hstar = (
                 hfact
                 * (
                     p_plasma_loss_mw
@@ -952,7 +957,7 @@ class PlasmaConfinementTime(Model):
         elif (
             self.data.physics.i_rad_loss == ConfinementRadiationLossModel.FULL_RADIATION
         ):
-            self.data.physics.hstar = (
+            hstar = (
                 hfact
                 * (
                     p_plasma_loss_mw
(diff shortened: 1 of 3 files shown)


process/models/physics/confinement_time.py
```diff
@@ -1024,7 +1030,7 @@ def calculate_confinement_time(
         # Calculate H* non-radiation corrected H factor
         # Note: we will assume the IPB-98y2 scaling.
         if self.data.physics.i_rad_loss == 1:
```
chris-ashe: Can the `ConfinementRadiationLossModel` `enum` values be added here

process/models/physics/confinement_time.py
```diff
@@ -1049,7 +1055,7 @@ def calculate_confinement_time(
                 ** 0.31
```
chris-ashe, line 969: This will only work for the selected scaling if the loss power index is (1-0.31), so really just the IPB98 scalings in this case

chris-ashe (changes requested):
Just really needs a way to be able to take the loss power terms from whatever equation is being used as the value is hardcoded for a single scaling at the moment

process/models/physics/confinement_time.py
timothy-nunn: This will need to be turned into an issue. This PR is just stopping `hstar` being incorrectly set every time `calculate_confinement_time` is called

chris-ashe:
Will await for rebase to fix merge conflict before second review @timothy-nunn

timothy-nunn:
Ready @chris-ashe

chris-ashe (approved):

process/models/physics/confinement_time.py
```diff
@@ -952,7 +957,7 @@ def calculate_confinement_time(
         elif (
             self.data.physics.i_rad_loss == ConfinementRadiationLossModel.FULL_RADIATION
```
clmould, line 958: Should the `elif` statements here also use `ConfinementRadiationLossModel(self.data.physics.i_rad_loss)` instead of just `self.data.physics.i_rad_loss`?

process/models/physics/confinement_time.py
clmould: Just in response to this - has an issue been made? Can it be linked here please

process/models/physics/confinement_time.py
timothy-nunn: The whole point of doing `ConfinementRadiationLossModel(self.data.physics.i_rad_loss)` is that an error will be raised if `self.data.physics.i_rad_loss` is not in the enum (ie its validation). It is only necessary to do this in the first `if` because this will catch and throw and any validation error.

process/models/physics/confinement_time.py
timothy-nunn: #4443

process/models/physics/confinement_time.py
clmould: Aah I see, that makes sense

clmould (approved):

PR #4405, opened 2026-07-02
Clean up `process_output`

timothy-nunn:
Remove legacy handling of bytes (required with the Fortran interface)

Update PROCESS to only use `ovarre` where able to.

### DIFF
--- a/process/core/process_output.py
+++ b/process/core/process_output.py
@@ -1,14 +1,20 @@
+"""Handles writing output to PROCESS MFile/OUTFile."""
+
 from contextlib import suppress
 from pathlib import Path
 
 import numpy as np
 
 from process.core import constants
+from process.core.exceptions import ProcessValueError
 
 
 class OutputFileManager:
+    """Manages the opening of regular/idempotence output files."""
+
     @classmethod
     def open_files(cls, output_prefix: str, *, mode="w"):
+        """Setup the handlers for MFile and OUTFile for writing."""
         cls._outfile = open(  # noqa: SIM115
             Path(output_prefix + "OUT.DAT"), mode
         )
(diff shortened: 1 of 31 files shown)


process/core/process_output.py
```diff
@@ -38,17 +48,28 @@ def close_idempotence_files(cls, output_prefix: str):
...
+        The file is not recognised as an MFile, OUTFile, or terminal.
+    """
     if file == constants.MFILE:
         OutputFileManager._mfile.write(f"{string}\n")
     elif file == constants.NOUT:
         OutputFileManager._outfile.write(f"{string}\n")
     elif file == constants.IOTTY:
         print(string)
+    else:
+        error_msg = f"Unknown file identifier {file}."
```
clmould: Would it be worth also including "The file is not recognised as an MFile, OUTFile, or terminal." here after the current error message so it is clearer for the user if this error occurs?

process/core/process_output.py
clmould, line 217: Can `ocosts` be removed too?

process/core/process_output.py
```diff
@@ -203,13 +222,5 @@ def ovarrf(file, descr: str, varnam: str, value, output_flag: str = ""):
     ovarre(file, descr, varnam, value, output_flag)
```
clmould, line 222: Think this needs deleting too (if I try and suggest the change it keeps not working/formatting weirdly so having to comment this!)

clmould (changes requested):

process/core/process_output.py
timothy-nunn: You'll notice that ocosts has a different order of arguments so it can be replaced, but not easily. It is also used by some other packages.

process/core/process_output.py
timothy-nunn: I would like to but need to work with @j-a-foster  first to update some external packages that still use `ovarrf`

clmould: Okay makes sense! I have made a 'blocked' issue for that #4439 so that can be done at a later date when possible

clmould (approved):

PR #4299, opened 2026-06-08
Update some constraints vars

chris-ashe:
This pull request focuses on standardizing and updating variable names across the codebase and input files for improved clarity and consistency, especially for constraint and physics-related parameters. It also updates documentation to reflect these new names and ensures all example input files use the revised identifiers.

**Variable Renaming and Consistency**

* Updated variable names in `process/core/input.py` and all example input files to use clearer, more descriptive identifiers for key constraints and physics parameters, such as:
    - `fdene` → `f_nd_plasma_electron_limit_max`
    - `fiooic` → `f_j_tf_wp_critical_max`
    - `psepbqarmax` → `p_div_bt_q_aspect_rmajor_max_mw`
    - `nflutfmax` → `flu_tf_neutron_fast_max`
    - `f_alpha_energy_confinement_min` → `f_t_alpha_energy_confinement_min`
    - `zeff_max` → `n_charge_plasma_effective_vol_avg_max`
    - `aplasmin` -> `rminor_min`
    - `avail_min` -> `f_t_plant_available_min`
    - `fradpwr` -> `f_p_plasma_seperatrix_rad_max`
    - `pseprmax` -> `p_plasma_separatrix_rmajor_max_mw`

**Documentation Updates**

* Updated all relevant documentation files to reference the new variable names and provide accurate descriptions of their usage and default values, ensuring alignment with code changes. 

**Example Input File Modernization**

* Refactored all example input files to use the new variable names, ensuring they are compatible with the updated code and documentation. 

**Minor Improvements**

* Clarified comments and docstrings for improved readability in the codebase.

These changes collectively improve the maintainability and clarity of the codebase by enforcing consistent naming conventions and ensuring that documentation and examples are up to date.

## Checklist

I confirm that I have completed the following checks:

- [ ] My changes follow the [PROCESS style guide](https://ukaea.github.io/PROCESS/development/standards/)
- [ ] I have justified any large differences in the regression tests caused by this pull request in the comments.
- [ ] I have added new tests where appropriate for the changes I have made.
- [ ] If I have had to change any existing unit or integration tests, I have justified this change in the pull request comments.
- [ ] If I have made documentation changes, I have checked they render correctly.
- [ ] I have added documentation for my change, if appropriate.

PR #4406, opened 2026-07-02
Fix Stellarator regression failures 

timothy-nunn:
Reverts constraint 17 to pre-#4299. 

The issue is that constraint 17 used to be formatted as https://github.com/ukaea/PROCESS/blob/a68bb7f9af7774ad4b0cd5663bc9fa576c429ee5/process/core/solver/constraints.py#L594-L605

Importantly, it used `pden_plasma_rad_mw`. Now, the constraint uses `f_p_plasma_separatrix_rad` which is calculated as
https://github.com/ukaea/PROCESS/blob/024a2f2e5a9ac750132f6ed010cbe03a34351916/process/models/physics/physics.py#L1037-L1042

However this uses `p_plasma_rad_mw` which also includes the SOL radiation

https://github.com/ukaea/PROCESS/blob/a68bb7f9af7774ad4b0cd5663bc9fa576c429ee5/process/models/stellarator/stellarator.py#L2170

which is never accounted for in pden_plasma_rad_mw. This changed the constraint, making the problem harder to solve, and led to constant failures when running the `stellarator-helias` test.

The diffs in the `stellarator-helias` test should revert it back to [this file](https://github.com/timothy-nunn/process-tracking-data/blob/master/stellarator_helias_MFILE_a68bb7f9af7774ad4b0cd5663bc9fa576c429ee5.DAT).

Thanks @clmould for identifying constraint 17 was the problem!

### DIFF
--- a/process/core/solver/constraints.py
+++ b/process/core/solver/constraints.py
@@ -592,8 +592,17 @@ def constraint_equation_17(constraint_registration, data):
     f_p_plasma_separatrix_rad_max: maximum allowed plasma radiation fraction at the
     separatrix
     """
+    if data.stellarator.istell != 0:
+        # Remove the contribution of psolradmw from f_p_plasma_separatrix_rad
+        # TODO: this is replicating behaviour before #4299
+        # is this really what should happen?
+        f_rad_sol = data.physics.psolradmw / data.physics.p_plasma_heating_total_mw
+        f_p_plasma_separatrix_rad = data.physics.f_p_plasma_separatrix_rad - f_rad_sol
+    else:
+        f_p_plasma_separatrix_rad = data.physics.f_p_plasma_separatrix_rad
+
     return leq(
-        data.physics.f_p_plasma_separatrix_rad,
+        f_p_plasma_separatrix_rad,
         data.constraints.f_p_plasma_separatrix_rad_max,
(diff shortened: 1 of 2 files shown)


process/models/physics/confinement_time.py
```diff
@@ -968,6 +968,10 @@ def calculate_confinement_time(
             eden_plasma_electrons_thermal_vol_avg / t_electron_energy_confinement
         ) / 1e6  # Convert from W/m³ to MW/m³
 
+        # import ipdb
```
chris-ashe: Accidentally left in?

process/models/physics/confinement_time.py
```diff
@@ -968,6 +968,10 @@ def calculate_confinement_time(
...
+
+        # ipdb.set_trace()
```
chris-ashe: Same here

chris-ashe (changes requested):

process/models/physics/confinement_time.py
timothy-nunn: Yes :D

chris-ashe (approved):

chris-ashe (approved):

clmould (approved):

PR #4415, opened 2026-07-06
Fix PLR0911 ruff errors

timothy-nunn:
Rewrite `parameter_type` for IN.DAT reading so that each variable is parsed by a dedicated function. This function can then be reused for the array parsing too which reduces the amount of return statements.


je-cook (approved):
thanks!

PR #4440, opened 2026-07-15
Increase stellarator-helias epsfcn

timothy-nunn:
At the moment `stellarator-helias` does not converge within its first 100 iterations. PROCESS then automatically revises the epsfcn (step length for calculating derivatives) from `1e-3` to `1e-2`. Instead of waiting for this to happen automatically, I propose the input file is updated to reflect this. 

@jjwalkowiak what do you think?

### DIFF
--- a/tests/regression/input_files/stellarator_helias.IN.DAT
+++ b/tests/regression/input_files/stellarator_helias.IN.DAT
@@ -227,7 +227,7 @@ f_nd_impurity_electrons(14) = 1.0E-5       *Tungsten
 ioptimz  = 1            *Code operation switch (1: Optimisation, VMCON only)
 maxcal   = 100          *Maximum number of VMCON iterations
 minmax   = 6            *Switch for figure-of-merit (7: Min Capital Cost)
-epsfcn   = 0.01
+epsfcn   = 0.01         *Default for this parameter is 1e-3 however this specific IN.DAT and starting point benefit from a larger step length
 runtitle = SQuID
 
 *-----------------Tfcoil Variables-----------------*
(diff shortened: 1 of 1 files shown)


jjwalkowiak:
My only concern would be that someone can try to use this regression test as template for other study and then start with 1e-2 as default option. But if you add a comment in the input file which explains that usually we start with 1e-3, then it is fine.
There might be also some issue with the results. I think it might be slightly different when you start directly from 1e-2 However I am not sure how it exactly works. Does the new epsfcn inherits the result from the last iteration of the previous epsfcn as a starting point, or it gets back to the defaults?

timothy-nunn:
> My only concern would be that someone can try to use this regression test as template for other study and then start with 1e-2 as default option. But if you add a comment in the input file which explains that usually we start with 1e-3, then it is fine. There might be also some issue with the results. I think it might be slightly different when you start directly from 1e-2 However I am not sure how it exactly works. Does the new epsfcn inherits the result from the last iteration of the previous epsfcn as a starting point, or it gets back to the defaults?

Thanks for the reply @jjwalkowiak. I will add a comment. I suspect that some of the state remains over from the previous failed optimisation hence why the regression tests have changed. However, this optima will be just as valid as the previous solution, it is probably just closer to the initial point.

Looking at the two solutions:
Old (`main`) objective value (cost of electricity): `1.21492798588871253e+02`
New (this PR) objective value: `1.21493505404321311e+02 `

This is so small that it does not even show up in the `0.2%` difference regression report. 


The only large changes to the solution are constraint residuals. These changing indicate that a different solution is being found, however this solution has near identical objective value and all the constraints are being satisfied.

clmould:
The reasoning for this change makes sense to me, so if @jjwalkowiak is happy I can merge it :)

timothy-nunn:
Was also thinking if changing the step length feels a little bit hacky we can always try and find a different starting point for the test instead.

jjwalkowiak:
I think it is fine to just change step length in that case. Normally I don't change starting step length, because I don't know in advance which will be the best for convergence. But as it is a testcase and we now exactly what we expect, I would say it is fine as long as everyone knows that a real study should have different setup.

PR #4468, opened 2026-07-25
Add AvailabilityModel enum for plant availability models

chris-ashe:
## Checklist

### DIFF
--- a/process/core/scan.py
+++ b/process/core/scan.py
@@ -17,6 +17,7 @@ from process.core.solver import constraints
 from process.core.solver.solver_handler import SolverHandler
 from process.data_structure.numerics import FiguresOfMerit, PROCESSRunMode
 from process.data_structure.scan_variables import IPNSCNS, NOUTVARS, ScanData
+from process.models.availability import AvailabilityModel
 
 if TYPE_CHECKING:
     from process.core.model import DataStructure, Model
@@ -1145,7 +1146,10 @@ class Scan:
             case 20:
                 self.data.constraints.t_burn_min = swp[iscn - 1]
             case 22:
-                if self.data.costs.i_plant_availability == 1:
+                if (
+                    AvailabilityModel(self.data.costs.i_plant_availability)
+                    == AvailabilityModel.WARD_TAYLOR
+                ):
                     raise ProcessValueError(
                         "Do not scan f_t_plant_available if i_plant_availability=1"
                     )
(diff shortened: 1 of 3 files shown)


timothy-nunn (changes requested):
I believe you are still missing some places where the switch value is checked:

Image


Image

process/core/scan.py
```diff
@@ -1145,7 +1146,10 @@ def scan_select(self, nwp, swp, iscn):
             case 20:
                 self.data.constraints.t_burn_min = swp[iscn - 1]
             case 22:
-                if self.data.costs.i_plant_availability == 1:
```
timothy-nunn, line 1148: Is there a reason this only applies to `i_plant_availability == 1`? I find it more likely that this should be checking `AvailabilityModel(self.data.costs.i_plant_availability) != AvailabilityModel.INPUT`... I.e. don't scan if `f_t_plant_available` is not being input because the scanned value will get overwritten by the model

process/core/solver/objectives.py
```diff
@@ -70,7 +71,10 @@ def objective_function(minmax: int, data: DataStructure) -> float:
     elif figure_of_merit == FiguresOfMerit.PULSE_LENGTH:
         objective_metric = data.times.t_plant_pulse_burn / 2.0e4
     elif figure_of_merit == FiguresOfMerit.PLANT_AVAILABILITY_FACTOR:
-        if data.costs.i_plant_availability != 1:
+        if (
+            AvailabilityModel(data.costs.i_plant_availability)
+            != AvailabilityModel.WARD_TAYLOR
+        ):
```
timothy-nunn, line 77: Similar question, is this correct anymore. Shouldn't you be able to optimise if using model 1, 2, or 3?

timothy-nunn (commented):
Couple of questions about code that seems might be outdated and from when we only had 1 availability model. Otherwise happy with code changes you have made.

process/core/scan.py
chris-ashe: Yes thats true will change the above

process/core/scan.py
```diff
@@ -1148,7 +1148,7 @@ def scan_select(self, nwp, swp, iscn):
             case 22:
                 if (
                     AvailabilityModel(self.data.costs.i_plant_availability)
-                    == AvailabilityModel.WARD_TAYLOR
+                    != AvailabilityModel.USER_INPUT
                 ):
                     raise ProcessValueError(
                         "Do not scan f_t_plant_available if i_plant_availability=1"
```
timothy-nunn: This message needs to be updated to reflect the new error condition

timothy-nunn (changes requested):

timothy-nunn (approved):

PR #4482, opened 2026-07-28
Scan improvements

je-cook:
## Description

This is part of #4308 but starting to split out things that can be reviewed now. 
This removes a lot of duplication around scans.

Also there scan variable 14 was not accessible before (i have made it so here). Is that wanted?

## Checklist

- [x] My changes follow the [PROCESS style guide](https://ukaea.github.io/PROCESS/development/standards/)


process/core/scan.py
```diff
@@ -21,169 +25,196 @@
...
+        var_area = getattr(data, self.area.value)
+
+        if self.number == 22 and var_area.i_plant_availability == 1:
+            raise ProcessValueError(
+                "Do not scan f_t_plant_available if i_plant_availability=1"
+            )
+
+        if "__" in self.name:
+            name, index = self.name.split("__")
+            getattr(var_area, name)[int(index)] = sweep_val
```
timothy-nunn:
```suggestion
            getattr(var_area, name)[int(index) - 1] = sweep_val
```

For example `boundu10` (scan 13) used to read
```python
case 13:
    self.data.numerics.boundu[9] = swp[iscn - 1]
```

process/core/scan.py
```diff
@@ -21,169 +25,196 @@
...
 
 @dataclass
 class ScanVariable:
-    variable_name: str
-    variable_description: str
-    variable_num: int
+    """Scan variable container"""
+
+    number: int
+    area: SVE = field(repr=False)
```
timothy-nunn: I'm not a _massive_ fan of this name. Maybe a docstring to make it clear that this is essential `data_structure.area`

process/core/scan.py
```diff
@@ -21,169 +25,196 @@
...
+        ValueError
+            description not set
+        """
+        if hasattr(self, "_description_"):
+            return self._description_
+        raise ValueError("Data not available")
+
+    def get_val(self, mfile, scan):
+        """Get value from mfile"""
+        # TODO this might fail for array variables eg boundu(10)?
```
timothy-nunn: The problem here is not that it is an array variable (there are no such things as array variables in an MFile

For example, the following will work
```python
mfile.get("f_nd_impurity_electrons(13)")
```

because it exists in the mfile:
```
Xe_concentration_________________________________________________________ (f_nd_impurity_electrons(13))__ 6.56757143700323126e-04 
```

However the boundu/boundl variables are not written out to the MFile

timothy-nunn: Maybe we should write the scan value to the MFile at the end of each scan point that way we don't rely on the variable being written the MFile with the same name as the scan variable

process/core/scan.py
```diff
@@ -21,169 +25,196 @@
...
+        """Variable description
+
+        Raises
+        ------
+        ValueError
+            description not set
+        """
+        if hasattr(self, "_description_"):
+            return self._description_
+        raise ValueError("Data not available")
```
timothy-nunn:
```suggestion
        raise ValueError("Description not available")
```

timothy-nunn: I don't know whether you think these should be `ProcessValueError`... I'm undecided either way

timothy-nunn (changes requested):
GitHub won't show me `process/core/io/plot/scans.py` and tbh I've tested that the plotting scripts work as before so I might not both doing a detailed review of it.

process/core/scan.py
je-cook: Yeah this is lazyness on my part will improve. 

Some future ideas:
We could also move this over to data structure in some form to shorten some of the really long `self.data.<area>....` maybe. You could then if we add a mechanic just do `self.TF` or whatever the shorthand is.

process/core/scan.py
je-cook: https://github.com/ukaea/PROCESS/issues/4487

timothy-nunn (commented):
Before:
Image

After:
Image

If we are going to include the variable name instead of the Latex symbol, can we make sure it considers your additional name field. Here the x variable is `b_tf_inboard_peak_symmetric`

je-cook:
thats a bug, I think fixed

timothy-nunn (approved):

PR #4483, opened 2026-07-29
Add pressure peaking fraction value

chris-ashe:
This pull request introduces several enhancements to how plasma pressure and temperature profiles are calculated, stored, and reported, particularly improving the handling of pedestal profile cases. The main changes include adding new derived quantities (such as central-to-volume-averaged ratios), calculating and outputting pedestal and separatrix pressures, and updating the summary plots and output files to reflect these new metrics.

**Enhancements to plasma profile calculations:**

- Added calculation and storage of the ratio of central to volume-averaged plasma electron temperature (`f_temp_plasma_electron_on_axis_vol_avg`) and plasma thermal pressure (`f_pres_plasma_thermal_on_axis_vol_avg`). These are now computed in both parabolic and pedestal parameterisations, and in the profile factors calculation. 
- For pedestal profile cases, the code now calculates and stores the plasma thermal pressure at the pedestal (`pres_plasma_pedestal_thermal`) and at the separatrix (`pres_plasma_separatrix_thermal`). 
**Output and reporting improvements:**

- The output file generation now includes the new ratio quantities and, for pedestal profiles, the pedestal and separatrix pressures. 
- 🐛 Fixed units in the output for plasma thermal energies from `[W]` to `[J]` for clarity and correctness.

**Plotting and summary display updates:**

- The plasma summary plots now display the new ratios and, when appropriate, show pedestal and separatrix pressures in the plot annotations. 

**Codebase and import updates:**

- Added necessary imports for `PlasmaProfileShapeType` to support the new logic for pedestal profile handling.

These changes improve the physical fidelity and reporting of plasma profile data, especially for cases with pedestal structure, and enhance the usability of the output for analysis and validation.## Description

## Checklist

### DIFF
--- a/process/data_structure/physics_variables.py
+++ b/process/data_structure/physics_variables.py
@@ -1055,6 +1055,12 @@ class PhysicsData:
     pres_plasma_fuel_profile: list[float] = field(default_factory=list)
     """Profile of fuel pressure in plasma [Pa]"""
 
+    pres_plasma_pedestal_thermal: float = 0.0
+    """Plasma pedestal thermal pressure (p_ped) (no fast ions or beam pressure) [Pa]"""
+
+    pres_plasma_separatrix_thermal: float = 0.0
+    """Plasma separatrix thermal pressure (pₛₑₚ) (no fast ions or beam pressure) [Pa]"""
+
     j_plasma_on_axis: float = 0.0
     """Central plasma current density (j₀) [A/m²]"""
 
--- a/process/data_structure/physics_variables.py
+++ b/process/data_structure/physics_variables.py
@@ -1040,6 +1040,9 @@ class PhysicsData:
     pres_plasma_thermal_on_axis: float = 0.0
     """Plasma central thermal pressure (p₀) (no fast ions or beam pressure) [Pa]"""
 
+    f_pres_plasma_thermal_on_axis_vol_avg: float = 0.0
+    """Ratio of central plasma thermal pressure to volume averaged (p₀/⟨p⟩)"""
+
     pres_plasma_thermal_total_profile: list[float] = field(default_factory=list)
     """Profile of total pressure in plasma [Pa]"""
 
--- a/process/data_structure/physics_variables.py
+++ b/process/data_structure/physics_variables.py
@@ -1403,6 +1403,9 @@ class PhysicsData:
     temp_plasma_electron_on_axis_kev: float = 0.0
     """Plasma central electron temperature (Tₑ₀) [keV]"""
 
+    f_temp_plasma_electron__on_axis_vol_avg: float = 0.0
+    """Ratio of plasma central electron temperature to volume averaged (Tₑ₀/⟨Tₑ⟩)"""
+
     temp_plasma_electron_density_weighted_kev: float = 0.0
     """Density weighted average electron temperature (⟨Tₑ⟩_n) [keV]"""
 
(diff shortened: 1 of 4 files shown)


process/data_structure/physics_variables.py
```diff
@@ -992,6 +992,9 @@ class PhysicsData:
     pres_plasma_thermal_on_axis: float = 0.0
     """Plasma central thermal pressure (p₀) (no fast ions or beam pressure) [Pa]"""
 
+    f_pres_plasma_thermal_on_axis_vol_avg: float = 0.0
```
grmtrkngtn, line 995: Maybe this is a real nit-pick but if it's a ratio should we use the "fraction" prefix for the variable name?

mkovari:
Hi Chris, Could you quote the actual equations from Johner that you have used, or your own derivations?  For example in `def tcore()`, I can see that the result arises from the integral in Appendix C, but the actual derivation of the function is not too obvious.

process/data_structure/physics_variables.py
chris-ashe: We dont really have anything else in the style guide that fits the bill. `f_` still represents a dimensionless unit that is the portion of something else

process/models/physics/plasma_profiles.py
```diff
@@ -284,6 +294,33 @@ def calculate_profile_factors(self):
...
+            # Find rho index closest to the normalized pedestal positions
+            pedestal_rho = np.min([
+                self.data.physics.radius_plasma_pedestal_temp_norm,
+                self.data.physics.radius_plasma_pedestal_density_norm,
+            ])
+            closest_idx = np.argmin(np.abs(rho - pedestal_rho))
+            mask = np.zeros_like(rho, dtype=bool)
+            mask[closest_idx:] = True
+            dpres_drho_pedestal = dpres_drho[mask]
+            max_grad_idx_pedestal = np.argmax(dpres_drho_pedestal)
```
grmtrkngtn, line 316: Should this use np.argmin? Since the pressure decreases with increasing radius, the steepest pedestal gradient should be the most negative value

chris-ashe:
> Hi Chris, Could you quote the actual equations from Johner that you have used, or your own derivations? For example in `def tcore()`, I can see that the result arises from the integral in Appendix C, but the actual derivation of the function is not too obvious.

Hi @mkovari if this is for the pedestal pressure I have done a very quick estimate via the first derivative at the pedestal point which is normally the local minimum. Though I am thinking if we are able to just take the arithmetic mean of the two pedestal positions for density and temperature and then take a value at that index of the pedestal profile. Though I am open to sugestions on how to calculate. An analytical expression would always be better in this case

process/models/physics/plasma_profiles.py
chris-ashe: I found that using `np.argmin` gets you closer to the pedestal in the large tokamak case as it plateus very quickly before the pedestal point. If you go for the max at the pedestal drop their is about a 10kPa difference. I do worry that this will break if a very flat profile is used so the pressure profile is practically vertical before the pedestal point

image

grmtrkngtn: Ok, thanks this picture was helpful.

I also checked this out myself, and you're right for this type of profile argmax picks out a better point but I also worry this method is not going to be so robust over different profiles.

grmtrkngtn: I can't come up with something which seems to work better however.

grmtrkngtn (approved):
I think the changes are good. The only thing is that I wonder how robustly we can pick out the pedestals across different profiles.

je-cook (approved):

je-cook (approved):

PR #4491, opened 2026-07-30
Some linting improvments to water and vacuum

je-cook:
## Description

This fixes all linting problems on water_use and fixes some on vacuum while trying to avoid changes that #4493 is already aiming to accomplish 

## Checklist

### DIFF
--- a/process/models/vacuum.py
+++ b/process/models/vacuum.py
@@ -54,46 +54,42 @@ class Vacuum(Model):
             * constants.UMASS
         )
 
-        self.i_vacuum_pumping = self.data.vacuum.i_vacuum_pumping
+        vp = self.data.vacuum
+        bld = self.data.build
+        phy = self.data.physics
 
-        # i_vacuum_pumping required to be compared to a b string
-        # as this is what f2py returns
-        if self.i_vacuum_pumping == "old":
+        if vp.i_vacuum_pumping == "old":
             (
                 pumpn,
-                self.data.vacuum.n_vv_vacuum_ducts,
-                self.data.vacuum.dlscal,
-                self.data.vacuum.m_vv_vacuum_duct_shield,
-                self.data.vacuum.dia_vv_vacuum_ducts,
+                vp.n_vv_vacuum_ducts,
+                vp.dlscal,
+                vp.m_vv_vacuum_duct_shield,
+                vp.dia_vv_vacuum_ducts,
             ) = self.vacuum(
-                self.data.physics.p_fusion_total_mw,
-                self.data.physics.rmajor,
-                self.data.physics.rminor,
-                0.5e0
-                * (
-                    self.data.build.dr_fw_plasma_gap_inboard
-                    + self.data.build.dr_fw_plasma_gap_outboard
-                ),
-                self.data.physics.a_plasma_surface,
-                self.data.physics.vol_plasma,
-                self.data.build.dr_shld_outboard,
-                self.data.build.dr_shld_inboard,
-                self.data.build.dr_tf_inboard,
-                self.data.build.r_shld_inboard_inner
-                - self.data.build.dr_shld_vv_gap_inboard
-                - self.data.build.dr_vv_inboard,
+                phy.p_fusion_total_mw,
+                phy.rmajor,
+                phy.rminor,
+                0.5e0 * (bld.dr_fw_plasma_gap_inboard + bld.dr_fw_plasma_gap_outboard),
+                phy.a_plasma_surface,
+                phy.vol_plasma,
+                bld.dr_shld_outboard,
+                bld.dr_shld_inboard,
+                bld.dr_tf_inboard,
+                bld.r_shld_inboard_inner
+                - bld.dr_shld_vv_gap_inboard
+                - bld.dr_vv_inboard,
                 self.data.tfcoil.n_tf_coils,
                 self.data.times.t_plant_pulse_dwell,
-                self.data.physics.nd_plasma_electrons_vol_avg,
+                phy.nd_plasma_electrons_vol_avg,
                 self.data.divertor.n_divertors,
                 qtorus,
                 gasld,
                 output=output,
             )
             # MDK pumpn is real: convert to integer by rounding.
-            self.data.vacuum.n_vac_pumps_high = math.floor(pumpn + 0.5e0)
-        elif self.i_vacuum_pumping == "simple":
-            self.data.vacuum.n_iter_vacuum_pumps = self.vacuum_simple(output=output)
+            vp.n_vac_pumps_high = math.floor(pumpn + 0.5e0)
+        elif vp.i_vacuum_pumping == "simple":
+            vp.n_iter_vacuum_pumps = self.vacuum_simple(output=output)
         else:
             logger.error(
                 f"i_vacuum_pumping is invalid: {self.data.vacuum.i_vacuum_pumping}"
@@ -157,7 +153,7 @@ class Vacuum(Model):
             self.outfile,
             "Switch for vacuum pumping model",
             "(i_vacuum_pumping)",
-            f'"{self.i_vacuum_pumping}"',
+            f'"{self.data.vacuum.i_vacuum_pumping}"',
         )
         process_output.ocmmnt(
             self.outfile,
--- a/process/models/vacuum.py
+++ b/process/models/vacuum.py
@@ -175,7 +175,7 @@ class Vacuum(Model):
         )
         process_output.ocmmnt(
             self.outfile,
-            "(50 Pa m3 s-1 = 1.2e+22 molecules/s), all operating at the same time",
+            " (50 Pa m3 s-1 = 1.2e+22 molecules/s), all operating at the same time",
         )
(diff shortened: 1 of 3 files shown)


process/models/water_use.py
```diff
@@ -305,3 +275,56 @@ def cooling_water_body(self, wastetherm: float, output: bool):
                 self.data.water_use.wateruseonethru,
                 "OP ",
             )
+
+
+@dataclass
+class CoolingWaterBodyCoeffs:
+    """Cooling water body coefficients"""
+
+    a: float
```
clmould, line 283: Could some descriptions be added for these vars please? Or an issue made to do this?

timothy-nunn (commented):
Additional refactor that can be done to vacuum:
```python
self.i_vacuum_pumping = self.data.vacuum.i_vacuum_pumping

# i_vacuum_pumping required to be compared to a b string
# as this is what f2py returns
```

1. The assigning of `self.i_vacuum_pumping` is unnecessary, its use should be replaced with the `self.data.vacuum.i_vacuum_pumping`
2. The comment about "b string" is not correct nor needed since the move to the Python data structure.

process/models/water_use.py
```diff
@@ -139,122 +141,90 @@ def cooling_water_body(self, wastetherm: float, output: bool):
...
-            ) - (
-                0.611e0
-                * np.exp(
-                    (17.27e0 * self.data.water_use.watertemp)
-                    / (237.3e0 + self.data.water_use.watertemp)
+            # find evaporation ratio: ratio of the heat used to evaporate water
+            # to the total heat discharged through the tower
+            self.data.water_use.evapratio = (
+                icool.delta_e(
+                    self.data.water_use.windspeed,
```
clmould: Can these be keyword args please?

process/models/water_use.py
clmould: Ik this was already here, but can this be moved to constants.py instead plz?

process/models/vacuum.py
```diff
@@ -147,62 +143,61 @@ def vacuum_simple(self, output) -> float:
...
+            "OP ",
+        )
+        process_output.ocmmnt(
+            self.outfile, "Number of high vacuum pumps, each with the throughput"
+        )
+        process_output.ocmmnt(
+            self.outfile,
+            " of one ITER cryopump (50 Pa m3 s-1 = 1.2e+22 molecules/s),",
+        )
+        process_output.ovarre(
```
clmould, line 169: Ik this was already like this, but this line only reads correctly in the OUT.DAT - in the MFILE it makes no sense as the previous 2 comments are missing:

OUT.DAT:

Image

MFILE.DAT:

Image

I think either all 3 lines need outputting to the MFILE (but would be long), or only output a shorter sentence to the MFILE and keep the OUT.DAT the same?

(Can make an issue for this if that'd be better)

process/models/vacuum.py
clmould:
```suggestion
       ----------
       output :
           indicate whether output should be written to the output file, or not

       Returns
       -------
       npump:
           number of pumps for pumpdown and steady-state
```
Hopefully formats correctly! But looks like the output description was in the wrong place all this time

process/models/vacuum.py
```diff
@@ -534,195 +436,303 @@ def vacuum(
...
-            process_output.ocmmnt(
-                self.outfile, "The vacuum pumping system size is governed by the"
-            )
+    def _newton_method_duct_diameter(
+        self, d, i, s, xmult, l1, l2, l3, ntf, r0, aw, ritf, thcsh, ceff
+    ):
+        nflag = 0  # Control option if ducts are too small in x-sectional area
+        #  = 1 if problem is identified in output, but run continues
+        #  = 0 otherwise
+        #  Newton's method solution for duct diameter
```
clmould, line 467:
```suggestion
        
        #  Newton's method solution for duct diameter
```
Slightly clearer it isn't part of the comment above

process/models/water_use.py
```diff
@@ -305,3 +275,56 @@ def cooling_water_body(self, wastetherm: float, output: bool):
...
+            + (self.j * self.heatload**2)
+        )
+
+    def imp_heatload_conversion(self):
+        """Convert heat loading into cal/(cm2.sec)"""
+        return self.heatload * 1000000.0e0 * 0.239e0 / 40469000.0e0
+
+    def met_heatload_conversion(self):
+        """Convert heat loading to J/(m2.day)"""
+        return self.heatload * 1000000.0e0 / 4046.85642e0 * SECDAY
```
clmould: Do these all need to be of the form `x.ye0` still? Can the `e0` be removed, and any `.0`'s? (Also in other places in this file)

process/models/vacuum.py
```diff
@@ -534,195 +436,303 @@ def vacuum(
...
-                )
-            elif imax == 3:
-                process_output.ocmmnt(
-                    self.outfile, "requirements for helium ash removal."
-                )
             else:
-                process_output.ocmmnt(
-                    self.outfile, "requirements for D-T removal at fuelling rate."
+                logger.error(
+                    f"Newton's method not converging; check fusion power, te "
```
clmould:
```suggestion
                    f"Newton's method not converging; check p_fusion_total_mw and temp_plasma_electron_vol_avg_kev "
```

Probably will cause a line length failure sorry! Ik this was already like this, but `te` is in obsolete vars having been replaced with `temp_plasma_electron_vol_avg_kev` , and more consistent to either use var names for both (as ive suggested), or write both as words (if that makes sense) ?

process/models/vacuum.py
```diff
@@ -534,195 +436,303 @@ def vacuum(
...
+            process_output.ocmmnt(self.outfile, "Conductance is inadequate.")
+            process_output.oblnkl(self.outfile)
 
-            process_output.ovarre(
-                self.outfile, "Duct length, divertor to elbow (m)", "(l2)", l2, "OP "
-            )
-            process_output.ovarre(
-                self.outfile, "Duct length, elbow to pumps (m)", "(l3)", l3
+        i_fw_blkt_shared_coolant = (
+            "cryo "
```
clmould:
```suggestion
            "cryo"
```

clmould (changes requested):

process/models/vacuum.py
je-cook: I've reversed it so in the MFile the variable is first

process/models/water_use.py
je-cook: Issue please I have no idea what they are (they may be different for each case etc)

process/models/water_use.py
```diff
@@ -305,3 +274,54 @@ def cooling_water_body(self, wastetherm: float, output: bool):
...
+    g: float
+    h: float
+    i: float
+    j: float
+    heatload: float
+
+    def heat_ratio(self, watertemp, windspeed):
+        """Estimate how heat loading will raise temperature, for this water body"""
+        # convert self.data.water_use.windspeed to mph
+        windspeedmph = windspeed * 2.237e0
```
clmould:
```suggestion
        windspeedmph = windspeed * 2.237
```

process/models/water_use.py
clmould: Yea, there were no descriptions already here so no clue what they are! Made an issue: #4515

process/models/vacuum.py
```diff
@@ -147,62 +142,58 @@ def vacuum_simple(self, output) -> float:
...
+        process_output.ovarre(
+            self.outfile,
+            "Number of high vacuum pumps, each with the throughput of one ITER cryopump"
+            "(n_iter_vacuum_pumps)",
+            n_iter_vacuum_pumps,
+            "OP ",
+        )
+        process_output.ocmmnt(
+            self.outfile,
+            "(50 Pa m3 s-1 = 1.2e+22 molecules/s), all operating at the same time",
```
clmould:
```suggestion
            " (50 Pa m3 s-1 = 1.2e+22 molecules/s), all operating at the same time",
```

Just because before there was a space on the new line, I think to make it clearer it was continued from the line above

process/models/vacuum.py
```diff
@@ -534,195 +432,305 @@ def vacuum(
...
-                    self.outfile, "requirements for helium ash removal."
-                )
             else:
-                process_output.ocmmnt(
-                    self.outfile, "requirements for D-T removal at fuelling rate."
+                logger.error(
+                    "Newton's method not converging; check "
+                    "p_fusion_total_mw and temp_plasma_electron_vol_avg_kev "
+                    f"{self.data.physics.p_fusion_total_mw=} "
+                    f"{self.data.physics.temp_plasma_electron_vol_avg_kev=}"
```
clmould:
```suggestion
                    "p_fusion_total_mw and temp_plasma_electron_vol_avg_kev: "
                    f"{self.data.physics.p_fusion_total_mw=}, "
                    f"{self.data.physics.temp_plasma_electron_vol_avg_kev=}"
```

Super minor and ik it was already formatted that way, but just a tad easier to read with it split up with the : and , ?

clmould (changes requested):

clmould (approved):

PR #4493, opened 2026-07-31
Add vaccum pump dataclass

chris-ashe:
This pull request refactors the vacuum pump modeling in `process/models/vacuum.py` to use structured dataclasses for pump and gas species properties, replacing the previous use of raw lists and constants. This makes the code more maintainable, readable, and less error-prone. The main computational logic is updated to use these new dataclasses, improving clarity and extensibility.

**Vacuum pump and species modeling:**
* Introduced new frozen dataclasses: `VacuumSpecies` (to represent gas species parameters) and `VacuumPump` (as a base for pump specifications), along with concrete subclasses `TurbomolecularPump` and `CryoPump` for specific pump types. These replace hardcoded lists and constants for pump and species properties.
* The selection and use of pump specifications in the `vacuum` function now instantiate and use these dataclasses instead of raw lists.

**Computational logic updates:**
* Updated all calculations in the `vacuum` function to access pump speeds and conductance multipliers via the new dataclass structure, improving code clarity and reducing indexing errors. [[1]](diffhunk://#diff-35c2f5fd3bd05bc9eada48497db2b793c716dc32bcd84ba6082a3c5536dd2c24L406-R505) [[2]](diffhunk://#diff-35c2f5fd3bd05bc9eada48497db2b793c716dc32bcd84ba6082a3c5536dd2c24L443-R538) [[3]](diffhunk://#diff-35c2f5fd3bd05bc9eada48497db2b793c716dc32bcd84ba6082a3c5536dd2c24L507-R603)

**General improvements:**
* Added the necessary imports for `dataclasses` and related utilities to support the new structure.## Description

## Checklist

### DIFF
--- a/process/models/vacuum.py
+++ b/process/models/vacuum.py
@@ -2,7 +2,7 @@
 
 import logging
 import math
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, fields
 
 import numpy as np
 
@@ -45,6 +45,22 @@ class VacuumPump:
     """Multiplier to convert conductance from gas species i to nitrogen"""
     description: str = ""
     """Description of the vacuum pump"""
+
+    def species(self):
+        """Yield species name, pump speed, and conductance multiplier.
+
+        Yields
+        ------
+        tuple[str, float, float]
+            Species name, volumetric flow rate, and conductance multiplier.
+        """
+        for species_field in fields(self.volflow_pump):
+            species_name = species_field.name
+            yield (
+                species_name,
+                getattr(self.volflow_pump, species_name),
+                getattr(self.xmult, species_name),
+            )
 
 
 @dataclass
@@ -361,18 +377,15 @@ class Vacuum(Model):
         #  thickness
         thcsh = thshldi / 3.0e0
 
-        #  Multiplier to convert conductance from gas species i to nitrogen
-        xmult = [1.0e0, 0.423e0, 0.378e0, 0.423e0]
-        # nitrogen, D-T, helium, D-T again
         nduct = ntf * ndiv
 
         #  Speed of high-vacuum pumps (m^3/s)
         # nitrogen, DT, helium, DT again
-        sp = (
-            [1.95, 1.8, 1.8, 1.8]
+        pump = (
+            TurbomolecularPump()
             if VacuumPumpType(self.data.vacuum.i_vacuum_pump_type)
             == VacuumPumpType.TURBOMOLECULAR
-            else [9.0, 25.0, 5.0, 25.0]
+            else CryoPump()
         )
 
         #  Calculate required pumping speeds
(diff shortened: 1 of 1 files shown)


process/models/vacuum.py
```diff
@@ -15,6 +16,93 @@
...
+    name: str = "Cryopump"
+    volflow_pump: VacuumSpecies = field(
+        default_factory=lambda: VacuumSpecies(
+            nitrogen=9.0,
+            deuterium_tritium=25.0,
+            helium=5.0,
+            deuterium_tritium_again=25.0,
+        )
+    )
+    description: str = "Compound cryopump with nominal speed 10 m³/s"
```
timothy-nunn, line 103: Since these dataclasses are frozen couldn't we just make some instances of the base dataclass. E.g. 

```python
CRYO_PUMP = VacuumPump(
    "Cryopump", 
    VacuumSpecies(
        nitrogen=9.0,
        deuterium_tritium=25.0,
        helium=5.0,
        deuterium_tritium_again=25.0,
    ),
    "Compound cryopump with nominal speed 10 m³/s"
)
```

process/models/vacuum.py
```diff
@@ -15,6 +16,93 @@
...
+        tuple[str, float, float]
+            Species name, volumetric flow rate, and conductance multiplier.
+        """
+        for species_field in fields(self.volflow_pump):
+            species_name = species_field.name
+            yield (
+                species_name,
+                getattr(self.volflow_pump, species_name),
+                getattr(self.xmult, species_name),
+            )
```
timothy-nunn, line 63: I think that this `species` almost negates the use of a dataclass

process/models/vacuum.py
```diff
@@ -15,6 +16,93 @@
...
+    """Name of the vacuum pump type"""
+    volflow_pump: VacuumSpecies
+    """Volumetric flow rates of the vacuum pump for different gases in m³/s"""
+    xmult: VacuumSpecies = field(
+        default_factory=lambda: VacuumSpecies(
+            nitrogen=1.0e0,
+            deuterium_tritium=0.423e0,
+            helium=0.378e0,
+            deuterium_tritium_again=0.423e0,
+        )
```
timothy-nunn, line 43: Are these values constant? If so, this would make more sense as an enum

process/models/vacuum.py
chris-ashe: My only concern would be is if we ever need to add different pumps to different systems. We would then have the same class instance declared manually everywhere instead of just having a single cryopump definition that could be called

timothy-nunn: Because this class is frozen the same object could be used everywhere, right (it ~~cannot~~ should not be mutated)? Unless we are saying that the `volflow_pump` would be different in these different use cases?

Issue #4063, opened 2026-01-22
Testing of resistive TF coil model


jonmaddock:
The resistive TF coil model requires testing via integration and/or regression tests.

PR #4495, opened 2026-08-03
Fix  `dr_tf_wp_eff` so it can be non zero

### LINKED ISSUE #4039: dr_tf_wp_eff will always be zero
https://github.com/ukaea/PROCESS/blob/1c324336c6d10fc9b9ce5fa9acd996f8c62d364b/process/tf_coil.py#L3673C1-L3674C55

This calculation will always equal zero, the second term must be inner.

ajpearcey:
closes #4039

### DIFF
--- a/process/models/tfcoil/base.py
+++ b/process/models/tfcoil/base.py
@@ -2903,7 +2903,7 @@ class TFCoil(Model):
             r_wp_outer_eff = np.double(r_tf_wp_inboard_outer)
 
         # Thickness of the layer representing the WP in stress calculations [m]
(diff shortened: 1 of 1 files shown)


timothy-nunn (approved):

mkovari:
@timothy-nunn Do we have any tests for resistive coils?

timothy-nunn:
@mkovari we do not at the moment. But I have been keen for a while to design a new regression test that covers some of our as-yet untested models e.g. resistive coils, different blanket models, etc. I'll make an issue (if one does not exist) to do this.

timothy-nunn:
So there is an issue #4063 to add a new regression test that covers this model.

PR #4518, opened 2026-08-13
Add new quench protection constraint

ajpearcey:
Adds new quench protection metric.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,5 @@ env-fork
 !documentation/**/*.html
 process/_version.py
 *.ipynb
+cost_conf.toml
+
--- a/process/core/init.py
+++ b/process/core/init.py
@@ -995,11 +995,11 @@ def check_process(inputs, data):  # noqa: ARG001
                 + 4.0 * data.tfcoil.radius_cp_coolant_channel
             )
 
-        if data.numerics.boundl[139] < dr_tf_wp_min:
-            raise ProcessValidationError(
-                "The TF coil WP thickness (dr_tf_wp_with_insulation) must be at least",
-                dr_tf_wp_min=dr_tf_wp_min,
-            )
+        #if data.numerics.boundl[140] < dr_tf_wp_min:
+        #    raise ProcessValidationError(
+        #        "The TF coil WP thickness (dr_tf_wp_with_insulation) must be at least",
+        #        dr_tf_wp_min=dr_tf_wp_min,
+        #    )
 
     # Setting i_dx_tf_turn_general_input to true if dx_tf_turn_general is an input
     data.tfcoil.i_dx_tf_turn_general_input = abs(data.tfcoil.dx_tf_turn_general) > 0
--- a/process/core/solver/constraints.py
+++ b/process/core/solver/constraints.py
@@ -1963,6 +1963,7 @@ def constraint_equation_93(constraint_registration, data):
     len_tf_coil: TF coil circumference (m)
     n_tf_coil_turns: Number of turns per TF coil
     e_tf_magnetic_stored_total_gj: Total stored energy in TF coils (GJ)
+    magnetic_stored_energy_copper_vol_ratio: TF coil stored energy to copper volume ratio
     """
     vol_tf_copper = (
         data.tfcoil.f_a_tf_turn_cable_copper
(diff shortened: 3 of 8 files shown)


process/core/solver/constraints.py
```diff
@@ -1955,6 +1954,32 @@ def constraint_equation_92(constraint_registration, data):
     )
 
 
+@ConstraintManager.register_constraint(93, "GJ/m3", ">=")
```
chris-ashe, line 1957: Should this not be `<=` as it places an upper limit on the stored energy to copper ratio?

chris-ashe, line 1957:
```suggestion
@ConstraintManager.register_constraint(93, "GJ/m³", ">=")
```

process/core/solver/constraints.py
```diff
@@ -1955,6 +1954,32 @@ def constraint_equation_92(constraint_registration, data):
...
+    """Lower limit of the copper in TF coil needed for quench protection
+    f_a_tf_turn_cable_copper: copper fraction of cable conductor TF coils
+    dr_tf_turn: Turn radial dimension (m)
+    dx_tf_turn: Turn toroidal dimension (m)
+    len_tf_coil: TF coil circumference (m)
+    n_tf_coil_turns: Number of turns per TF coil
+    e_tf_magnetic_stored_total_gj: Total stored energy in TF coils (GJ)
+    magnetic_stored_energy_copper_vol_ratio: TF coil stored energy to copper volume ratio
+    """
+    vol_tf_copper = (
```
chris-ashe, line 1968: This value is going to be wildly different depending on what type of turn is used. `f_a_tf_turn_cable_copper` only represents the area fraction of copper in a superconducting strand in CICC configuration.

process/core/solver/constraints.py
```diff
@@ -1955,6 +1954,32 @@ def constraint_equation_92(constraint_registration, data):
...
+        data.tfcoil.f_a_tf_turn_cable_copper
+        * data.superconducting_tfcoil.dr_tf_turn
```
chris-ashe, line 1970: `a_tf_turn` could just be used here

process/core/init.py
```diff
@@ -996,11 +995,11 @@ def check_process(inputs, data):  # noqa: ARG001
                 + 4.0 * data.tfcoil.radius_cp_coolant_channel
             )
 
-        if data.numerics.boundl[139] < dr_tf_wp_min:
-            raise ProcessValidationError(
-                "The TF coil WP thickness (dr_tf_wp_with_insulation) must be at least",
-                dr_tf_wp_min=dr_tf_wp_min,
-            )
+        #if data.numerics.boundl[140] < dr_tf_wp_min:
```
chris-ashe, line 998: Accidentally commented out?

process/data_structure/constraint_variables.py
```diff
@@ -131,5 +131,8 @@ class ConstraintData:
     I.e. p_l_h_threshold_mw / p_plasma_separatrix_mw >= f_l_mode_margin
     """
 
+    magnetic_stored_energy_copper_vol_ratio: float = 0.25
```
chris-ashe, line 134: I am unsure about the definition of this and it doesn't follow the style guide. Does this mean 0.25 GJ of stored energy per 1 m^3 of copper in the TF coil is the upper limit?

chris-ashe (changes requested):
A few things about understanding of the constraint. Alot of the calcs need to be put into the specific TF configuration class

.gitignore
timothy-nunn, line 1: I tend to try and avoid personalised edits to the .gitignore as this makes it long and confusing

process/data_structure/constraint_variables.py
```diff
@@ -131,5 +131,8 @@ class ConstraintData:
...
+    """TF coil stored energy to copper volume ratio (`constraint equation 93`) """
```
timothy-nunn, line 135: Can we comment the units here, even if it is dimensionless `[-]`

process/data_structure/constraint_variables.py
ajpearcey: The otherway the ratio is the copper volume divided by stored energy.

process/data_structure/constraint_variables.py
ajpearcey: will add [m3/GJ]

process/core/solver/constraints.py
ajpearcey: Sure, but it is expressed in this constraint as Vol_Cu >=  stored energy. I think this is just down to how you want to write the constraint.

process/core/init.py
ajpearcey: I did comment this out, but I have a suspicion this just wrong. The original code checks agains ixc = 139, which is `dx_tf_hts_tape_copper` this doesn't match the error message or make sense why it would be checked. Also why check the lower bound? Whereas ixc = 140 is `dr_tf_wp_with_insulation`.

PR #4519, opened 2026-08-13
Updated PF numbering in outputs

### LINKED ISSUE #4502: OUT.DAT PF coil numbering
The OUT.DAT lists the PF coils starting from 0, but this should be updated to list them starting from 1:

Image

kj5248:
## Description

closes #4502 

In MFILE.DAT and OUT.DAT there were many places where PF coils were numbered from 0. This has been updated to start from 1.

Also in places such as Volt Second Consumption and Waveforms all the circuits were numbers, the final number has now shows CS as that's what it is. However in the PF Circuit Waveform Data section I left the circuit_var_name as the original as it is a var not an output name.

## Checklist

### DIFF
--- a/process/models/pfcoil.py
+++ b/process/models/pfcoil.py
@@ -2750,7 +2750,8 @@ class PFCoil(Model):
         inv_st_pulse = 1.0e0 / pf_d.f_j_cs_start_pulse_end_flat_top
 
         for k in range(self.data.pf_coil.n_pf_cs_plasma_circuits - 1):
-            if k == self.data.pf_coil.n_pf_cs_plasma_circuits - 2:
+            if (self.data.build.iohcl != 0) and (k ==
+                        self.data.pf_coil.n_pf_cs_plasma_circuits - 2):
                 line = "\tCS\t\t"
             else:
                 line = f"\t{k + 1}\t\t"
@@ -2768,7 +2769,8 @@ class PFCoil(Model):
         op.ocmmnt(self.outfile, "This consists of: CS coil field balancing:")
 
         for k in range(pf_d.n_pf_cs_plasma_circuits - 1):
-            if k == pf_d.n_pf_cs_plasma_circuits - 2:
+            if (self.data.build.iohcl != 0) and (k ==
+                                    self.data.pf_coil.n_pf_cs_plasma_circuits - 2):
                 op.write(
                     self.outfile,
                     (
@@ -2796,7 +2798,8 @@ class PFCoil(Model):
         op.oblnkl(self.outfile)
         op.ocmmnt(self.outfile, "And: equilibrium field:")
         for k in range(pf_d.n_pf_cs_plasma_circuits - 1):
-            if k == pf_d.n_pf_cs_plasma_circuits - 2:
+            if (self.data.build.iohcl != 0) and (k ==
+                                    self.data.pf_coil.n_pf_cs_plasma_circuits - 2):
                 op.write(
                     self.outfile,
                     (
@@ -2847,7 +2850,8 @@ class PFCoil(Model):
                 if k == self.data.pf_coil.n_pf_cs_plasma_circuits - 1:
                     circuit_name = f"Plasma Time point {jjj} (A)"
                     circuit_var_name = f"(plasmat{jjj})"
-                elif k == self.data.pf_coil.n_pf_cs_plasma_circuits - 2:
+                elif (self.data.build.iohcl != 0) and (k ==
+                                        self.data.pf_coil.n_pf_cs_plasma_circuits - 2):
                     circuit_name = f"CS Circuit Time point {jjj} (A)"
                     circuit_var_name = f"(cs t{jjj})"
                 else:
--- a/process/models/pfcoil.py
+++ b/process/models/pfcoil.py
@@ -2436,12 +2436,12 @@ class PFCoil(Model):
             "1.0e0",
         ])
 
-        for line in tabulate(
+        op.write(self.outfile, tabulate(
             pf_coil_geometry_rows,
             headers=["Coil", "R(m)", "Z(m)", "dR(m)", "dZ(m)", "turns"],
             tablefmt="plain",
-        ).splitlines():
-            op.write(self.outfile, line)
+            disable_numparse=True,
+        ))
 
         for k in range(pf_d.nef):
             op.ovarre(
@@ -2629,13 +2629,12 @@ class PFCoil(Model):
         ])
 
         op.oblnkl(self.outfile)
-        for line in tabulate(
+        op.write(self.outfile, tabulate(
             rows,
             headers=headers,
             tablefmt="plain",
             disable_numparse=True,
-        ).splitlines():
-            op.write(self.outfile, line)
+        ))
 
         op.oblnkl(self.outfile)
         op.ocmmnt(self.outfile, "----------------------------")
@@ -2683,13 +2682,12 @@ class PFCoil(Model):
         ))
 
         op.oblnkl(self.outfile)
-        for line in tabulate(
+        op.write(self.outfile, tabulate(
             rows,
             headers=headers,
             tablefmt="plain",
             disable_numparse=True,
-        ).splitlines():
-            op.write(self.outfile, line)
+        ))
         op.oblnkl(self.outfile)
         op.ovarre(
             self.outfile,
@@ -2707,7 +2705,6 @@ class PFCoil(Model):
         )
 
         op.osubhd(self.outfile, "Summary of volt-second consumption by circuit (Wb):")
-        op.write(self.outfile, "Circuit\t\t\tBOP\t\t\tBOF\t\tEOF")
         op.oblnkl(self.outfile)
         headers = [
             "Circuit",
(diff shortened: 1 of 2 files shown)


chris-ashe (changes requested):
@kj5248 Have noticed that the coil numbering in the plots now starts from 2. Can you quickly update that to and I will take a full look after

chris-ashe (changes requested):
All `OUT.DAT` looks ok, though I am getting a bug for `plot_pf_dimensions()` when running the ST regression were it does not do the final coil, Coil 8 in this case

kj5248:
@chris-ashe  > All `OUT.DAT` looks ok, though I am getting a bug for `plot_pf_dimensions()` when running the ST regression were it does not do the final coil, Coil 8 in this case

When I run the ST regression and use plot summary on it i get this which has 8 PF coils?

image

Although I did notice somewhere in the OUT.dat for ST which seems wrong but wanted to double check. Under waveforms it drops to 7 coils and a CS, feels like this is wrong?

image

chris-ashe (approved):

chris-ashe:
> @chris-ashe > All `OUT.DAT` looks ok, though I am getting a bug for `plot_pf_dimensions()` when running the ST regression were it does not do the final coil, Coil 8 in this case
> 
> When I run the ST regression and use plot summary on it i get this which has 8 PF coils?
> 
> image
> Although I did notice somewhere in the OUT.dat for ST which seems wrong but wanted to double check. Under waveforms it drops to 7 coils and a CS, feels like this is wrong?
> 
> image

Yeah, there must be a dodgy difference of 1 somewhere, the coil number should always be even. Also, if possible can you apply `tabulate` to the output to make everything line up properly, 2 birds with one stone

kj5248:
Tabulated the outputs. I also noticed what I am pretty sure is a duplicated output (it is in table form in the waveforms section) so I removed it @chris-ashe:

image

process/models/pfcoil.py
```diff
@@ -2728,17 +2756,26 @@ def outvolt(self):
...
+        for k in range(pulse_timings.n_pf_active_points_total):
+            line += [f"{pulse_timings.pf_active_cumulative[k]:.2f}"]
+        rows = [line]
+        for line in tabulate(
+            rows,
+            headers=headers[1:],
+            tablefmt="plain",
+            disable_numparse=True,
+        ).splitlines():
+            op.write(self.outfile, line)
```
je-cook: I dont think you need to loop over the lines I think you can write it directly as @timothy-nunn has done here https://github.com/ukaea/PROCESS/pull/4565/changes
```suggestion
        op.write(self.outfile, tabulate(
            rows,
            headers=headers[1:],
            tablefmt="plain",
            disable_numparse=True,
        ))
```

chris-ashe (changes requested):
The PF/CS mutual inductance matrix colour plot in `summary.py` is still starting from 0. Also the PF coil dimensions plot is not adding the dimensions for the final coil but the numbering is correct

chris-ashe (approved):

process/models/pfcoil.py
```diff
@@ -2660,26 +2665,39 @@ def outvolt(self):
...
-        op.write(
-            self.outfile,
-            f"CS coil:\t\t{pf.vs_cs_ramp:.2f}"
-            f"\t\t\t\t{pf.vs_cs_burn:.2f}\t\t\t{pf.vs_cs_total_pulse:.2f}",
-        )
-        op.write(
-            self.outfile, "\t" * 6 + "-" * 7 + "\t" * 4 + "-" * 7 + "\t" * 3 + "-" * 7
-        )
+        headers = ["", "Start-up (Vs)", "Burn (Vs)", "Total (Vs)"]
+        rows = []
```
clmould: TODO why ? would it work just declaring the list with entries ?

process/models/pfcoil.py
```diff
@@ -2728,17 +2760,28 @@ def outvolt(self):
...
-        line = "\t\t"
         for k in range(pulse_timings.n_pf_active_points_total):
             label = pulse_timings.POINT_ABBREVIATIONS[k]
-            line += f"\t\t{label}"
-        op.write(self.outfile, line)
+            headers.append(label)
+        line = []
+        for k in range(pulse_timings.n_pf_active_points_total):
+            line += [f"{pulse_timings.pf_active_cumulative[k]:.2f}"]
+        rows = [line]
```
clmould: TODO could this be combined with line = [] ?

process/models/pfcoil.py
```diff
@@ -2728,17 +2760,28 @@ def outvolt(self):
...
+            tabulate(
+                rows,
+                headers=headers[1:],
+                tablefmt="plain",
+                disable_numparse=True,
+            ),
+        )
+
+        op.oblnkl(self.outfile)
+        op.write(self.outfile, "Currents (A)")
```
clmould:
```suggestion
        op.write(self.outfile, "Currents [A]")
```

process/models/pfcoil.py
clmould:
```suggestion
        rows = []
        for k in range(pulse_timings.n_pf_active_points_total):
            rows += [f"{pulse_timings.pf_active_cumulative[k]:.2f}"]
```

I tested and this produces the same table, as `line` was only used to set `rows` so can use `rows` directly here

process/models/pfcoil.py
```diff
@@ -2660,26 +2665,39 @@ def outvolt(self):
...
-            self.outfile,
-            f"CS coil:\t\t{pf.vs_cs_ramp:.2f}"
-            f"\t\t\t\t{pf.vs_cs_burn:.2f}\t\t\t{pf.vs_cs_total_pulse:.2f}",
-        )
-        op.write(
-            self.outfile, "\t" * 6 + "-" * 7 + "\t" * 4 + "-" * 7 + "\t" * 3 + "-" * 7
-        )
+        headers = ["", "Start-up (Vs)", "Burn (Vs)", "Total (Vs)"]
+        rows = []
+        rows.extend((
```
clmould:
```suggestion
        rows = [
```

I think the `rows` can be immediately declared here instead of extending the array

process/models/pfcoil.py
```diff
@@ -2660,26 +2665,39 @@ def outvolt(self):
...
+                f"{pf.vs_cs_burn:.2f}",
+                f"{pf.vs_cs_total_pulse:.2f}",
+            ],
+            [
+                "Total",
+                f"{pf.vs_cs_pf_total_ramp:.2f}",
+                f"{pf.vs_cs_pf_total_burn:.2f}",
+                f"{pf.vs_cs_pf_total_pulse:.2f}",
+            ],
+        ))
```
clmould:
```suggestion
        ]
```

clmould (changes requested):
A couple minor things but happy the numbering now starts from 1 :)

process/models/pfcoil.py
kj5248: Found if I don't add the extra square brackets around rows here it produces:
image
```
rows = []
        for k in range(pulse_timings.n_pf_active_points_total):
            rows += [f"{pulse_timings.pf_active_cumulative[k]:.2f}"]
        rows = [rows]
```
This resolves it to show this:
image
If you can come up with a cleaner method than just adding extra brackets I'll change it.

process/models/pfcoil.py
```diff
@@ -2728,17 +2759,28 @@ def outvolt(self):
...
-        line = "\t\t"
         for k in range(pulse_timings.n_pf_active_points_total):
             label = pulse_timings.POINT_ABBREVIATIONS[k]
-            line += f"\t\t{label}"
-        op.write(self.outfile, line)
+            headers.append(label)
+        rows = []
+        for k in range(pulse_timings.n_pf_active_points_total):
+            rows += [f"{pulse_timings.pf_active_cumulative[k]:.2f}"]
+        rows = [rows]
```
clmould:
```suggestion
        pulse_times = []
        for k in range(pulse_timings.n_pf_active_points_total):
            pulse_times += [f"{pulse_timings.pf_active_cumulative[k]:.2f}"]
        rows = [pulse_times]
```

Sorry going back to your previous version as what I suggested wasn't super clear now since it didn't work! Perhaps `pulse_times` instead of `lines` as there's just 1 line in the table?

process/models/pfcoil.py
clmould: Sorry! I must've looked at the wrong bit of output when I thought I tested this!

process/models/pfcoil.py
```diff
@@ -2728,17 +2759,28 @@ def outvolt(self):
...
+        op.write(
+            self.outfile,
+            tabulate(
+                rows,
```
clmould:
```suggestion
        
        op.write(
            self.outfile,
            tabulate(
                [rows],
```

Could put the brackets inside the `tabulate` ? Though I'll leave this up to you, I wouldn't insist on this

process/models/pfcoil.py
clmould: I've taken a look and the only alternative I've found is to create a dictionary of the headers and corresponding times to use in `tabulate`, but I'm not sure it's any neater than the one set of extra brackets tbh!

process/models/pfcoil.py
```diff
@@ -2807,9 +2910,9 @@ def outvolt(self):
             "OP ",
         )
 
-        op.oshead(self.outfile, "PF Circuit Waveform Data")
+        # Add output for the MFILE to parse
```
clmould, line 2913:
```suggestion
        # Output to MFILE
```
Not sure the previous comment makes sense?

PR #4521, opened 2026-08-13
Expand fitted temperature range of impurities

chris-ashe:
## Description

Updates the loss function data for impurities to comprise of 600 elements from 0.25 eV up to 500 keV

## Checklist


process/data_structure/impurity_radiation_variables.py
```diff
@@ -54,7 +54,7 @@ class ImpurityRadiationData:
     )
 
     n_charge_impurity_profile: list[float] = field(
-        default_factory=lambda: np.zeros((N_IMPURITIES, 200))
+        default_factory=lambda: np.zeros((N_IMPURITIES, 600))
```
timothy-nunn: Could we make `600` a constant at the top of this file?

process/models/physics/impurity_radiation.py
timothy-nunn, line 1: We could then also use these constants here

timothy-nunn (changes requested):

timothy-nunn:
Requesting modeller review to confirm changes to unit tests and regression tests are acceptable

mkovari:
> Requesting modeller review to confirm changes to unit tests and regression tests are acceptable

As long as you can confirm you have checked that Lz and <Z> are the close to the old values for the old temperature range I don't there will be a problem.  

I find the high value of Lz for Fe below 1 eV surprising, but since we never use temperatures below 1 eV, and we never use Fe, this is hardly an issue.

timothy-nunn:
@mkovari here is a plot of the old data and new data (only the new data on the old time range). There is very little difference.
adas_radiation

mkovari (approved):

process/models/physics/impurity_radiation.py
```diff
@@ -31,7 +34,8 @@ def initialise_imprad(data: DataStructure):
     """
     errorflag = 0
 
-    table_length = 600  # Number of temperature and Lz values in data file
```
timothy-nunn: This is now unnecessary and `N_IMPURITIY_LOSS_FUNCTION_POINTS` can be used in place of `table_length`

timothy-nunn (changes requested):

process/models/physics/impurity_radiation.py
```diff
@@ -31,7 +34,8 @@ def initialise_imprad(data: DataStructure):
     """
     errorflag = 0
 
-    table_length = 200  # Number of temperature and Lz values in data file
+    # Number of temperature and Lz values in data file
+    table_length = N_IMPURITIY_LOSS_FUNCTION_POINTS
```
clmould: I agree with Tim's comment here - instead of having `table_length = N_IMPURITIY_LOSS_FUNCTION_POINTS` and then setting `len_tab=table_length` in all the calls to `init_imp_element()`, you can use `len_tab=N_IMPURITIY_LOSS_FUNCTION_POINTS` instead

documentation/scripts/plotting_scripts/impurity_loss_functions.py
```diff
@@ -0,0 +1,99 @@
...
+if __name__ == "__main__":
+    with resources.path(
+        "process.data.lz_non_corona_14_elements", "Ar_lz_tau.dat"
+    ) as imp_path:
+        data_folder = str(imp_path.parent) + "/"
+
+    if Path(data_folder).is_dir():
+        fig, ax = plt.subplots(figsize=(8, 6))
+        plot_line_brem_loss_function_profile(ax, impp=data_folder)
+    else:
```
clmould: I don't think this `else` will ever be hit. Ik this `if`/`else` comes from summary.py so I think this was just super outdated and doesn't work as intended any more:

`resources.path()` returns a `pathlib.Path` object, so `imp_path` is always a path to Ar_lz_tau.dat, and then `data_folder` is always pointing to the `lz_non_corona_14_elements` folder, so `Path(data_folder).is_dir()` is always `True`, so it will never hit this `else`

I tried to hit the else - by deleting the lz_non_corona_14_elements folder, deleting all the entries in this folder, deleting some of the entries in this folder - but always end up with a FileNotFoundError or ModuleNotFoundError.

I think that removing the `if`/`else` and instead just calling the plotting function here and in summary.py will work the same, and if any .dat files are missing then that will be covered by the FileNotFoundErrors

documentation/source/physics-models/plasma_radiation.md
```diff
@@ -32,19 +32,21 @@ $$
 where $P_i$ is the radiation per unit volume (excluding synchrotron radiation),
 $L_Z (Z_i, T)$ is the loss function for ion species $i$ at temperature $T$, and $n_i$ is the density of ion species $i$.
 
-The source data comprises of 200 separate data points with a fitted temperature range of $1 \ \text{eV}$ tp $40 \ \text{keV}$. The fitted radiation profiles for each species against temperature can be found in the figure below.
+The source data comprises of 600 separate data points with a fitted temperature range of $0.25 \ \text{eV}$ tp $500 \ \text{keV}$. The fitted radiation profiles for each species against temperature can be found in the figure below.
```
clmould:
```suggestion
The source data comprises of 600 separate data points with a fitted temperature range of $0.25 \ \text{eV}$ to $500 \ \text{keV}$. The fitted radiation profiles for each species against temperature can be found in the figure below.
```

clmould (changes requested):

process/core/io/plot/summary.py
```diff
@@ -16220,13 +16220,8 @@ def main_plot(
...
 
-    if Path(data_folder).is_dir():
-        imp = data_folder
-    else:
-        print(
-            "\033[91m Warning : Impossible to recover impurity data, try running the macro in the main/utility folder"
-        )
-        print("          -> No impurity plot done\033[0m")
+    imp = data_folder
+
```
clmould: It won't let me suggest it as a change, but instead of 
`data_folder = str(imp_path.parent) + "/"` then `imp = data_folder`, could it instead just be `imp = str(imp_path.parent) + "/"` ? Just removes an unnecessary re-assigning ? Could call it `imp_data_folder` if that would be clearer?

process/models/physics/impurity_radiation.py
```diff
@@ -325,10 +326,10 @@ def init_imp_element(
     )
     data.impurity_radiation.impurity_arr_len_tab[n_species_index - 1] = len_tab
 
-    if len_tab > 200:
+    if len_tab > N_IMPURITIY_LOSS_FUNCTION_POINTS:
```
clmould, line 329: Does this get hit any more? Since `len_tab` is always equal to `N_IMPURITIY_LOSS_FUNCTION_POINTS`?

clmould (changes requested):

process/models/physics/impurity_radiation.py
chris-ashe: In its current state no. If someone then goes and changes the assignment anywhere then this will catch it

clmould: Makes sense!

clmould (approved):

PR #4535, opened 2026-08-15
Reject zero-thickness superconducting TF at input validation

### LINKED ISSUE #4534: Superconducting TF silently loses all inboard thickness when dr_tf_wp_with_insulation is not an iteration variable
## Describe the bug
`Build.calculate_radial_build` only derives `dr_tf_inboard` from the winding-pack and case thicknesses when `dr_tf_wp_with_insulation` (ixc = 140) is an **iteration variable**:

```python
# Issue #514 Radial dimensions of inboard leg
# Calculate self.data.build.dr_tf_inboard if
# self.data.tfcoil.dr_tf_wp_with_insulation is an iteration variable (140)
if 140 in self.data.numerics.ixc[0 : self.data.numerics.n_iteration_variables]:
    self.data.build.dr_tf_inboard = (...)
```

If a user supplies `dr_tf_wp_with_insulation` as a plain input instead (a natural thing to do when freezing a design), and does not use ixc 13 or set `dr_tf_inboard` explicitly, `dr_tf_inboard` silently stays at its default of **0.0**: the TF coil vanishes from the radial build. The run then fails far downstream and confusingly — radial-build consistency (icc 11) violated by the full coil thickness, TF stress constraints evaluating at tens of GPa against an 850 MPa limit — with no message pointing at the cause. Whether a physical quantity is derived or defaulted-to-zero currently depends on the *numerics* configuration.

Reproduce: take `tests/regression/input_files/st_regression.IN.DAT`, replace `ixc = 140` with a fixed `dr_tf_wp_with_insulation = 0.63` (and `ixc = 57` with `dr_tf_nose_case = 0.25`); the radial build silently loses ~0.93 m and VMCON fails with no explanation.

## Proposed fix
Validate at input-check time (`check_process`) that a superconducting TF has a positive `dr_tf_inboard` when neither ixc = 13 nor ixc = 140 is active, with an actionable error message. Stellarators (which compute `dr_tf_inboard` during the model run) and IFE excluded. All seven shipped regression inputs pass the check. PR ready (found during an independent audit of v3.4.2; related to the older #514 TODO on making dr_tf_inboard an output).

dallonby:
Closes #4534

## Overview
Adds a `check_process` validation so that the silent zero-thickness-TF failure mode described in the issue becomes an immediate, actionable input error instead of an unexplained downstream solver failure.

## Changes
- `process/core/init.py`: after the existing ixc 13/140 mutual-exclusion check, raise `ProcessValidationError` when `i_tf_sup` is superconducting, neither ixc 13 nor ixc 140 is active, and `dr_tf_inboard <= 0`. The message tells the user the three ways to fix their file. Stellarators (`istell != 0`, which calculate `dr_tf_inboard` during the model run) and IFE are excluded.
- **New** `tests/unit/core/test_init.py`: five tests — the rejected configuration, plus accepted configurations for explicit thickness, ixc 140 active, resistive TF, and stellarator.
- `tests/unit/core/test_input.py` (test-suite change, per CONTRIBUTING): the parser tests run `init_process` on minimal input snippets (e.g. just `epsvmc = 1.0`) and relied on config validation never examining the TF geometry; their shared fixture scaffold now sets a valid `dr_tf_inboard` so the parser tests keep testing parsing. No expected values change.

## Behavioural impact
Valid configurations are unaffected (all seven shipped regression inputs pass: they each set ixc 13/140 or `dr_tf_inboard`, or are stellarator/IFE). The only newly-rejected configurations are ones that previously produced a machine with no inboard TF coil.

## Verification
- Full unit suite: 847 passed / 4 skipped; integration suite: 21 passed / 1 skipped (Python 3.12, local).

Found during an independent audit of v3.4.2.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/process/core/init.py
+++ b/process/core/init.py
@@ -289,6 +289,29 @@ def check_process(inputs, data):  # noqa: ARG001
             "Iteration variables 13 and 140 cannot be used simultaneously",
         )
 
+    # A superconducting TF coil must have a non-zero inboard thickness.
+    # dr_tf_inboard is only derived from the winding pack and case thicknesses
+    # when dr_tf_wp_with_insulation (ixc = 140) is an iteration variable (see
+    # Build.calculate_radial_build); with neither ixc = 13 nor ixc = 140 active
+    # it stays at its input value, and the default of 0 silently removes the
+    # TF coil from the radial build. Stellarators calculate dr_tf_inboard
+    # during the model run, so are excluded from this check.
+    if (
+        data.stellarator.istell == 0
+        and data.ife.ife == 0
+        and data.tfcoil.i_tf_sup == TFConductorModel.SUPERCONDUCTING
+        and not (data.numerics.ixc[: data.numerics.n_iteration_variables] == 13).any()
+        and not (data.numerics.ixc[: data.numerics.n_iteration_variables] == 140).any()
+        and data.build.dr_tf_inboard <= 0.0
+    ):
+        raise ProcessValidationError(
+            "dr_tf_inboard is not positive: the superconducting inboard TF coil"
+            " has no thickness. Set dr_tf_inboard (or use ixc = 13), or make"
+            " dr_tf_wp_with_insulation an iteration variable (ixc = 140) so that"
+            " dr_tf_inboard is derived from the winding pack and case thicknesses",
+            dr_tf_inboard=data.build.dr_tf_inboard,
+        )
+
     # Can't use c_tf_turn as iteration var, constraint or
     # input if i_tf_turns_integer == 1
     if (
--- a/process/core/init.py
+++ b/process/core/init.py
@@ -289,24 +289,28 @@ def check_process(inputs, data):  # noqa: ARG001
             "Iteration variables 13 and 140 cannot be used simultaneously",
         )
 
-    # A superconducting TF coil must have a non-zero inboard thickness.
-    # dr_tf_inboard is only derived from the winding pack and case thicknesses
-    # when dr_tf_wp_with_insulation (ixc = 140) is an iteration variable (see
-    # Build.calculate_radial_build); with neither ixc = 13 nor ixc = 140 active
-    # it stays at its input value, and the default of 0 silently removes the
-    # TF coil from the radial build. Stellarators calculate dr_tf_inboard
-    # during the model run, so are excluded from this check.
+    # An inboard TF coil must have a positive thickness. dr_tf_inboard is
+    # only derived from the winding pack and case thicknesses when
+    # dr_tf_wp_with_insulation (ixc = 140) is an iteration variable (see
+    # Build.calculate_radial_build); otherwise the input-file value is what
+    # the radial build uses, and the default of 0 silently removes the TF
+    # coil. This applies to superconducting and resistive conductors alike,
+    # and also when dr_tf_inboard itself is the iteration variable
+    # (ixc = 13): the input value seeds the first model evaluation, an
+    # exactly-zero value is only caught later by the generic
+    # iteration-variable check, and a negative value is not caught at all
+    # (the 1/value scaling in load_iteration_variables inverts the
+    # variable's bounds). Stellarators calculate dr_tf_inboard during the
+    # model run, so are excluded from this check.
     if (
         data.stellarator.istell == 0
         and data.ife.ife == 0
-        and data.tfcoil.i_tf_sup == TFConductorModel.SUPERCONDUCTING
-        and not (data.numerics.ixc[: data.numerics.n_iteration_variables] == 13).any()
         and not (data.numerics.ixc[: data.numerics.n_iteration_variables] == 140).any()
         and data.build.dr_tf_inboard <= 0.0
     ):
         raise ProcessValidationError(
-            "dr_tf_inboard is not positive: the superconducting inboard TF coil"
-            " has no thickness. Set dr_tf_inboard (or use ixc = 13), or make"
+            "dr_tf_inboard is not positive: the inboard TF coil has no"
+            " thickness. Set a positive dr_tf_inboard, or make"
             " dr_tf_wp_with_insulation an iteration variable (ixc = 140) so that"
             " dr_tf_inboard is derived from the winding pack and case thicknesses",
             dr_tf_inboard=data.build.dr_tf_inboard,
--- a/process/core/init.py
+++ b/process/core/init.py
@@ -287,6 +287,29 @@ def check_process(inputs, data):  # noqa: ARG001
             "Iteration variables 13 and 140 cannot be used simultaneously",
         )
 
+    # A superconducting TF coil must have a non-zero inboard thickness.
+    # dr_tf_inboard is only derived from the winding pack and case thicknesses
+    # when dr_tf_wp_with_insulation (ixc = 140) is an iteration variable (see
+    # Build.calculate_radial_build); with neither ixc = 13 nor ixc = 140 active
+    # it stays at its input value, and the default of 0 silently removes the
+    # TF coil from the radial build. Stellarators calculate dr_tf_inboard
+    # during the model run, so are excluded from this check.
+    if (
+        data.stellarator.istell == 0
+        and data.ife.ife == 0
+        and data.tfcoil.i_tf_sup == TFConductorModel.SUPERCONDUCTING
+        and not (data.numerics.ixc[: data.numerics.n_iteration_variables] == 13).any()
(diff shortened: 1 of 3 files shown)


je-cook:
@ukaea/process-model-review this may adversely effect the optimiser as it is currently allowed to search invalid regions of space I think?

I suppose this is only checking init

mkovari:
Obviously this needs to be fixed - maybe someone who has looked at this code more recently than me - @chris-ashe or someone?

dallonby:
Init only. `check_process` is called from exactly one place, `init_process` (`init.py:79`), which `SingleRun.__init__` runs once (`main.py:339` then `:432`) before the `Models` objects are even built (`main.py:340`), so before `Scan`, `SolverHandler` or VMCON get anywhere near it. Scans don't re-init between points (`scan.py` just sets the swept variable), VaryRun spawns a fresh `SingleRun` per iteration, and `dr_tf_inboard` only gets written during model evaluation when ixc 140 is active (`build.py:1684`) or for stellarators, both of which the check excludes.

To be sure I wasn't reasoning my way past something, I instrumented it on the branch: 1 call to `check_process` vs 26 model evaluations (spherical_tokamak_eval, fsolve), 2029 (large_tokamak_nof, VMCON), 4286 (low_aspect_ratio_DEMO) and 16175 (a 3-point scan). One call every time, before the solve. So the optimiser's freedom to wander through invalid regions is untouched; the check only looks at inputs that are constant for the whole run.

Two things I should say explicitly though. It is a deliberate behaviour change for scaffold inputs that rely on the `dr_tf_inboard = 0` default, which is why the `test_input.py` fixture needed a value. For what it's worth, on `main` those inputs currently either hit an uncaught `ZeroDivisionError` (`tfcoil/base.py:424`, circular case) or, with `i_tf_case_geom = 1`, report a "feasible solution" with a winding pack of -0.34 m and a negative turn count. And I only covered the superconducting case that was reported; a resistive TF with the same zero goes down the same `build.py` path, so it could be extended in a follow-up if you'd like.

process/core/init.py
```diff
@@ -287,6 +287,29 @@ def check_process(inputs, data):  # noqa: ARG001
...
+    # when dr_tf_wp_with_insulation (ixc = 140) is an iteration variable (see
+    # Build.calculate_radial_build); with neither ixc = 13 nor ixc = 140 active
+    # it stays at its input value, and the default of 0 silently removes the
+    # TF coil from the radial build. Stellarators calculate dr_tf_inboard
+    # during the model run, so are excluded from this check.
+    if (
+        data.stellarator.istell == 0
+        and data.ife.ife == 0
+        and data.tfcoil.i_tf_sup == TFConductorModel.SUPERCONDUCTING
+        and not (data.numerics.ixc[: data.numerics.n_iteration_variables] == 13).any()
```
timothy-nunn: I think this condition should be removed because on the first iteration with `ixc = 13` the IN.DAT-defined value of `dr_tf_inboard` is used which would cause the model errors this condition is designed to protect against.

timothy-nunn: This would also require a slight change to the error message below.

process/core/init.py
```diff
@@ -287,6 +287,29 @@ def check_process(inputs, data):  # noqa: ARG001
...
+    # dr_tf_inboard is only derived from the winding pack and case thicknesses
+    # when dr_tf_wp_with_insulation (ixc = 140) is an iteration variable (see
+    # Build.calculate_radial_build); with neither ixc = 13 nor ixc = 140 active
+    # it stays at its input value, and the default of 0 silently removes the
+    # TF coil from the radial build. Stellarators calculate dr_tf_inboard
+    # during the model run, so are excluded from this check.
+    if (
+        data.stellarator.istell == 0
+        and data.ife.ife == 0
+        and data.tfcoil.i_tf_sup == TFConductorModel.SUPERCONDUCTING
```
timothy-nunn: Hopefully someone from @ukaea/process-model-review can confirm, but I think we can probably test this condition for superconducting _and_ resistive TF coils. Resistive TF still uses `dr_tf_inboard` and it would not surprise me if it being 0 or negative would be equally problematic here too.

tests/unit/core/test_input.py
```diff
@@ -12,7 +12,12 @@
 
 @pytest.fixture
 def data_structure_obj():
-    return DataStructure()
+    data = DataStructure()
+    # These parser tests run init_process on minimal input snippets; give the
+    # scaffold a valid TF thickness so configuration validation (which rejects
+    # a zero-thickness superconducting TF) does not reject the scaffold.
+    data.build.dr_tf_inboard = 1.0
+    return data
```
timothy-nunn: I want to avoid having to do specific setup in this test wherever possible. A better option for me is to turn off `check_process` since its not actually useful for this test.

```python
@pytest.fixture(autouse=True)
def turn_off_check_process(monkeypatch):
    monkeypatch.setattr(init, "check_process", lambda *_: None)
```

Rebasing to include #4553 (now merged onto main) will also be useful because that will check none of these tests fail because of the above change.

timothy-nunn (changes requested):

timothy-nunn:
@je-cook the only impact on the "solver" (on the solution) will be that some users may need to update their initial point if they currently have `dr_tf_inboard /= 0` which could change the path/solution.

dallonby:
Thanks — all four points taken, in a89dcb28 (rebased onto main, so #4553's `match=` checks run against the fixture change).

One precision on the ixc = 13 case, because it slightly strengthens your change: with `dr_tf_inboard = 0` and ixc = 13, the run already failed before any model evaluation — `load_iteration_variables` computes `scale = 1/value`, and `check_iteration_variable` rejects an exactly-zero value with the generic "iteration variable is 0" error. What the exemption really left open is a *negative* input value: that passes `check_iteration_variable` (it only tests |v| ≤ 1e-12/NaN/inf), the negative scale inverts the variable's scaled bounds, and the first model evaluation runs with a negative thickness. The check now rejects both at input validation with the actionable message; the message no longer suggests ixc = 13 as a remedy.

The conductor-model condition is removed rather than widened: `dr_tf_inboard` is only derived under ixc = 140 (`Build.calculate_radial_build`) or for stellarators, neither of which depends on `i_tf_sup`, so resistive coils had the identical path. Your `check_process` monkeypatch is in as written.

Unit 855 passed / 4 skipped; integration 21 passed / 1 skipped — all shipped inputs pass the widened check.

tests/unit/core/test_init.py
```diff
@@ -0,0 +1,103 @@
...
+    assert "dr_tf_inboard" not in _validation_error_message(data)
+
+
+def test_thickness_iteration_variable_does_not_exempt():
+    """ixc = 13 does not exempt a non-positive input value: it seeds the
+    first model evaluation, an exactly-zero value is only rejected later by
+    the generic iteration-variable check, and a negative value is not
+    rejected at all (the 1/value scaling inverts the variable's bounds).
+    """
+    for bad_value in (0.0, -0.5):
```
timothy-nunn: Sorry that I missed this in my first review, I would prefer that this was `parametrize`'d rather than using a `for` loop over bad values

tests/unit/core/test_input.py
```diff
@@ -10,6 +10,12 @@
 from process.data_structure.numerics import PROCESSRunMode
 
 
+@pytest.fixture(autouse=True)
+def turn_off_check_process(monkeypatch):
+    """These are parser tests; configuration validation is not under test."""
```
timothy-nunn:
```suggestion
    """These are parser tests; configuration validation is not a part of test."""
```

timothy-nunn (changes requested):

timothy-nunn:
Thanks @dallonby, I think I was actually wrong about the ixc = 13 case because the input parser would reject a negative value. Nevertheless I still think keeping the validation logic as is makes sense as it does no harm.

tests/unit/core/test_init.py
```diff
@@ -0,0 +1,105 @@
+"""Unit tests for input sanity checks in process.core.init.check_process."""
+
+import pytest
+
+from process.core.exceptions import ProcessValidationError
+from process.core.init import check_process
+from process.core.model import DataStructure
```
timothy-nunn:
```suggestion
from process.core.data_structure.base import DataStructure
```

A little change we made yesterday that has not quite filtered through

timothy-nunn (commented):
Just a change we made yesterday that has not quite filtered through. Otherwise all looks good, thank you for finding this bug and providing a fix!

timothy-nunn (approved):

Issue #3968, opened 2025-11-04
:bug:Handling of radiation in the power balance - question


mkovari:
Jon has suggested there is a very large error in the calculation of power balance.

Investigation:

The [MFILE](https://github.com/ukaea/PROCESS/blob/0fdad95a6100f3719f95cdbd624b869ef1bf7ad3/examples/data/large_tokamak_1_MFILE.DAT#L1134) still contains the power balance across the separatrix, and it still balances:

```
 Transport_power_from_scaling_law_(MW)________________________________ (pscalingmw)__________________    2.9785E+02
 Radiation_power_from_inside_"radius_plasma_core_norm"_(MW)___________(pcoreradmw.)_________________    9.1650E+01
 Total_(MW)__________________________________________________________________________________________   3.8950E+02
 Alpha_power_deposited_in_plasma_(MW)_________________________________ (falpha*palpmw)______________    3.0746E+02 
 Power_from_charged_products_of_DD_and/or_D-He3_fusion_(MW)___________ (pchargemw.)_________________    1.2659E+00   
 Ohmic_heating_(MW)___________________________________________________ (p_plasma_ohmic_mw.)_____________6.2284E-01
 Injected_power_deposited_in_plasma_(MW)_____________________________ (p_hcd_injected_total_mw)________ 8.0143E+01
 Total_(MW)__________________________________________________________________________________________   3.8950E+02
```
(Note that `pcoreradmw` has been renamed as `p_plasma_inner_rad_mw` in the code but not in the MFILE.)

This is based on the assumption that the “transport power” derived from the energy confinement scaling law includes (despite its name) the radiation power from outside the “core radius”.

<img width="394" height="399" alt="Image" src="https://github.com/user-attachments/assets/bb1587af-2e9c-4f31-b85a-e977357e2cf5" />

In other words, radiation from the outer part of the plasma actually reduces the power transported by charged particles across the separatrix by the same amount, because they have to _add up to_ a fixed amount.  Whether this is true is a big question, but it is what we have claimed since 2014.  It is also in the docs.

<img width="2071" height="1204" alt="Image" src="https://github.com/user-attachments/assets/0dfed066-9cc5-4207-8559-de2f57e78236" />

## Tracing the 'transport power' `scalingmw` _backwards_ through the code.  ##
(I have removed the class names `physics_variables` , `imp_rad` and `self` for clarity).
```Python
        # Total transport power from scaling law (MW)
[...]

PR #4536, opened 2026-08-16
Impurity radiation corrections

jonmaddock:
Demonstrates the effect of various options for impurity radiation and allows their plotting: this is not intended to be merged in its current form, but as an investigation. Primarily, this demonstrates the importance of separating the reduced core radiation used for the confinement time calculation and the total radiation considered in the plasma power balance, which I currently believe is incorrect.

I hope the following figures highlight a few things and justify the actual changes I want to make. They are all made with the DEMO LAR solution point and varying tungsten fractions. Deliberate horizontal "jitter" is applied to the scatter plots to ensure overlapping points are discernable.

## Impurity radiation profiles
rad_profiles
The current impurity power densities are incorrect because they are multiplied prematurely by the normalised minor radius $\rho$ in preparation for the power integral. This has the effect of the power density dropping to 0 for $\rho = 0$, which is incorrect. The core radiation reduction is also evident outside of the confinement time scaling calculation, where it is only intended to be used. This is corrected. The edge integral is now performed explicitly in the 4th subfigure, which is used later subsequently.

## Radiation power
rad_totals

When the core radiation reduction is removed, the core radiation obviously increases, but the edge radiation decreases, due to the "edge = total - core" calculation that is currently used. The edge is optionally now integrated explicitly. The total radiation is the same in all cases.

## Separate radiation reduction for confinement time scaling only
pden_core_tauE

Two plasma core radiation power densities are now calculated: one for the confinement time scaling only, the other for the rest of the code. They only differ when the `tau E rad reduction` option is selected (brown), which is the only difference between the plots. This separates the radiation calculated for the rest of the models from the reduced value required by the scaling.

## Effect on plasma power balance (PPB)
ppb

The effect of these various options on the absolute plasma power balance constraint residual in MW (i.e. MW m^-3,  then multiplied by plasma volume ) is shown. The "original" point being slightly above zero is probably due to using a slightly out-of-date solution vector relative to the current Process version. One real difference occurs from using the edge radiation in the PPB (purple) which should always be done in my view: the plasma is not in thermal equilibrium if the edge radiation is ignored. The second, larger, difference is due to the use of the reduced core radiation in the confinement time calculation only, rather than globally (i.e. in the PPB too). This is how this "correction" was originally intended to be used I believe: as a way of effectively changing the confinement time scaling for highly-radiative regimes. It should not affect the actual calculated radiation elsewhere in the models. Around 100 MW for low tungsten values is fairly significant.

I would like to always use the un-reduced total radiation (core and edge) in the plasma power balance calculation, and the reduced core radiation in the confinement time calculation only. This would mean that `i_rad_loss` and `f_p_plasma_core_rad_reduction` would only apply to the the confinement time calculation. I'm keen to hear your thoughts.

### DIFF
--- a/process/core/caller.py
+++ b/process/core/caller.py
@@ -419,7 +419,7 @@ def finalise(models, data, ifail: int, non_idempotent_msg: str | None = None):
         po.oheadr(constants.NOUT, "Final UNFEASIBLE Point")
 
     # Output relevant to no optimisation
-    if data.numerics.ioptimz == PROCESSRunMode.EVALUATION:
+    if data.numerics.ioptimz in {PROCESSRunMode.EVALUATION, PROCESSRunMode.SOLUTION}:
         output_evaluation(data)
 
     # Print non-idempotence warning to OUT.DAT only
--- a/process/core/caller.py
+++ b/process/core/caller.py
@@ -206,9 +206,28 @@ class Caller:
                     OutputFileManager.close_idempotence_files(
                         self.data.globals.output_prefix
                     )
+
+                    # Now idempotent, return
+                    # Pass model caller and opt params for stability constraint evaluation
+                    con_residuals_normalised, _, con_residuals, _, _ = (
+                        constraints.constraint_eqns(
+                            self.data.numerics.neqns + self.data.numerics.nineqns,
+                            -1,
+                            self.data,
+                        )
+                    )
+                    # Evaluate constraints and store in numerics during solver iterations:
+                    # can be used in objective function 20. Hence evaluate before objective
+                    # calculated
+                    self.data.numerics.constraint_residuals_normalised = (
+                        con_residuals_normalised
+                    )
+                    self.data.numerics.constraint_residuals = con_residuals
+                    # Evaluate objective function and constraints
+                    objf = objective_function(self.data.numerics.minmax, self.data)
                     # Write final output file and mfile
                     finalise(self.models, self.data, ifail)
-                    return
+                    return objf, con_residuals_normalised
 
                 # Mfiles not yet idempotent: need to re-evaluate models
                 logger.debug("Mfiles not idempotent, evaluating models again")
--- a/process/data_structure/numerics.py
+++ b/process/data_structure/numerics.py
@@ -49,13 +49,15 @@ class PROCESSRunMode(IntEnum):
     """In this mode, the code will not perform any optimisation, and will instead
     simply evaluate the constraints for the given input parameters, which is useful
     for testing and for evaluating the performance of a given design point without
-    trying to optimise it. Internally, PROCESS uses `fsolve` (a Newton-Krylov/hybrd
+    trying to optimise it.
+    """
+    SOLUTION = (-1, "Solution mode (no optimisation)")
+    """Internally, PROCESS uses `fsolve` (a Newton-Krylov/hybrd
     root-finding method from `scipy.optimize`) to seek a *consistent* solution by
     varying a subset of the iteration variables until the consistency constraints
     (equality constraints whose residuals must be driven to zero) are simultaneously
     satisfied; no figure-of-merit is optimised, and the solver simply tries to find
-    a root of the constraint-residual vector.
-    """
+    a root of the constraint-residual vector."""
     OPTIMISATION = (1, "Optimisation mode (e.g. via VMCON)")
     """In this mode, the code will perform optimisation using the VMCON solver
     (or a custom solver if specified) to try to find a design point that optimises
--- a/process/data_structure/numerics.py
+++ b/process/data_structure/numerics.py
@@ -184,6 +184,14 @@ class NumericsData:
     nvar: int = 0
     """number of iteration variables to use"""
 
+    # Constraint residuals, updated on every iteration
+    constraint_residuals_normalised: list[float] = field(
+        default_factory=lambda: np.array([0] * IPEQNS)
+    )
+    constraint_residuals: list[float] = field(
+        default_factory=lambda: np.array([0] * IPEQNS)
+    )
+
     nviter: int = 0
     """number of optimisation iterations performed"""
 
--- a/process/data_structure/physics_variables.py
+++ b/process/data_structure/physics_variables.py
@@ -1078,6 +1078,9 @@ class PhysicsData:
     pden_plasma_core_rad_mw: float = 0.0
     """total core radiation power per volume (MW/m3)"""
 
+    pden_plasma_core_rad_tauE_mw: float = 0.0
+    """reduced total core radiation power for tauE calculation (MW/m3)"""
+
     p_dd_total_mw: float = 0.0
     """deuterium-deuterium fusion power (MW)"""
 
--- a/process/main.py
+++ b/process/main.py
@@ -119,6 +119,8 @@ from process.models.tfcoil.superconducting import (
 )
 from process.models.vacuum import Vacuum, VacuumVessel
 from process.models.water_use import WaterUse
+from process.core.caller import write_output_files
+from process.core.solver.iteration_variables import load_iteration_variables
 
 PACKAGE_LOGGING = True
 """Can be set False to disable package-level logging, e.g. in the test suite"""
@@ -449,10 +451,19 @@ class SingleRun:
         # ioptimz == 1: optimisation
         if self.data.numerics.ioptimz == PROCESSRunMode.OPTIMISATION:
             pass
-        elif self.data.numerics.ioptimz == PROCESSRunMode.EVALUATION:
-            # No optimisation:
-            # solve equality (consistency) constraints only using fsolve (HYBRD)
+        # ioptimz == -1: solution
+        elif self.data.numerics.ioptimz == PROCESSRunMode.SOLUTION:
+            # Solve equality (consistency) constraints only using fsolve (HYBRD)
             self.solver = "fsolve"
+        # ioptimz == -2: evaluation
+        elif self.data.numerics.ioptimz == PROCESSRunMode.EVALUATION:
+            # Evalutation only: compute the output variables now
+            # Get optimisation parameters x, evaluate models
+            load_iteration_variables(self.data)
+            self.ifail = 6
+            write_output_files(data=self.data, models=self.models, ifail=self.ifail)
+            self.show_errors()
+            return
         else:
             raise ValueError(
                 f"Invalid ioptimz value: {self.data.numerics.ioptimz}. Please "
--- a/process/models/costs/costs.py
+++ b/process/models/costs/costs.py
@@ -2854,7 +2854,8 @@ class Costs(Model):
 
         #  Capital recovery factor
 
-        crfcdr = (fefcdr * self.data.costs.discount_rate) / (fefcdr - 1.0e0)
+        # crfcdr = (fefcdr * self.data.costs.discount_rate) / (fefcdr - 1.0e0)
+        crfcdr = 1.0
 
         #  Annual cost of replacements
 
--- a/process/models/physics/confinement_time.py
+++ b/process/models/physics/confinement_time.py
@@ -18,6 +18,7 @@ from process.data_structure.physics_variables import (
     PlasmaIgnitionModel,
 )
 from process.models.physics.plasma_geometry import PlasmaGeom
+from process.models.physics import impurity_radiation
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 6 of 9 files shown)


jonmaddock:
This appears to independently resolve https://github.com/ukaea/PROCESS/issues/3968#issuecomment-3491154712.

mkovari:
> this is not intended to be merged in its current form, but as an investigation.

Can you clarify this?

je-cook:
> > this is not intended to be merged in its current form, but as an investigation.
> 
> Can you clarify this?

I think he means it should probably be split into a few separate PRs after the initial review discussion. Also I guess the switches wont be needed if we agree on the solution

mkovari:
> This appears to independently resolve [#3968 (comment)](https://github.com/ukaea/PROCESS/issues/3968#issuecomment-3491154712).

This link is to the comment on volume-integration.  Was that intended?

process/models/physics/impurity_radiation.py
```diff
@@ -23,6 +23,8 @@
 
 logger = logging.getLogger(__name__)
 include_edge_radiation = False
+int_edge_rad = False
```
grmtrkngtn: This is not used.

mkovari:
> I think he means it should probably be split into a few separate PRs after the initial review discussion. Also I guess the switches wont be needed if we agree on the solution

So do we actually need to approve the PR?

je-cook:
no but a review is needed of the content

grmtrkngtn:
## Impurity radiation profiles

I agree with the proposed changes to the impurity radiation profiles. Moving the (\rho) weighting from the stored power-density profile into the volume integral is a clear-cut fix.

## Core radiation reduction

I also agree with separating the physical core radiation from the reduced value used in the confinement-time calculation. Retaining the option to apply the reduction to the scaling law only seems like a reasonable compromise but the physical power balance must use the full radiation.

## Outstanding

The corrected confinement calculation still relies on two input assumptions: `radius_plasma_core_norm` and 
`f_p_plasma_core_rad_reduction`.

I'm not sure we still still need the latter, why would we ever not subtract the full radiation (why would we use the factor?)

EUROfusion colleagues have also suggested that the effective core boundary could be calculated from the radial power flow rather than prescribed as an input. My understanding is that METIS calculates the cumulative net power, heating minus local radiation, as a function of radius and uses its maximum to determine the effective transport-driving power. Should we investigate this here, or would that be better treated as a separate follow-up to this accounting PR?

grmtrkngtn (commented):
## Impurity radiation profiles

I agree with the proposed changes to the impurity radiation profiles. Moving the (\rho) weighting from the stored power-density profile into the volume integral is a clear-cut fix.

## Core radiation reduction

I also agree with separating the physical core radiation from the reduced value used in the confinement-time calculation. Retaining the option to apply the reduction to the scaling law only seems like a reasonable compromise but the physical power balance must use the full radiation.

## Outstanding

The corrected confinement calculation still relies on two input assumptions: `radius_plasma_core_norm` and 
`f_p_plasma_core_rad_reduction`.

I'm not sure we still still need the latter, why would we ever not subtract the full radiation (why would we use the factor?)

EUROfusion colleagues have also suggested that the effective core boundary could be calculated from the radial power flow rather than prescribed as an input. My understanding is that METIS calculates the cumulative net power, heating minus local radiation, as a function of radius and uses its maximum to determine the effective transport-driving power. Should we investigate this here, or would that be better treated as a separate follow-up to this accounting PR?

jonmaddock:
@mkovari these changes solve https://github.com/ukaea/PROCESS/issues/3968 (the issue, not just the comment within the issue) as it separates the power density calculation and its integral. This draft PR is intended to show how the above plots were generated to better understand the problem due to the confusion of multiple issues acting together and the overall effect of the changes I intend to make, which will be a separate PR. These changes are described in the final paragraph of the PR description.

@grmtrkngtn 

> The corrected confinement calculation still relies on two input assumptions: radius_plasma_core_norm and
f_p_plasma_core_rad_reduction.
I'm not sure we still still need the latter, why would we ever not subtract the full radiation (why would we use the factor?)

That is how the confinement time scaling law has effectively been modified (according to Lux et. al.) to model the highly-radiative regime: of the core region (`radius_plasma_core_norm`) only use some fraction of the radiation (`f_p_plasma_core_rad_reduction`). Changing this would require performing Lux et. al.'s analysis again to achieve a different way of modifying the scaling to fit highly-radiative data, which is outside the scope of this task. I'm just interested in using the correct amount of radiation in the plasma power balance calculation here.

Any calculation of the core boundary is an additional feature.

I'd be grateful for a review from @mkovari , @chris-ashe and @ajpearcey .

mkovari (approved):
I can't say I have read it all but I am keen to see what this looks like when merged into main.

ajpearcey:
I would be interested to see how the changed PPB changes the final solution vector for the DEMO-LAR. How big is this shift in the Psep value? Also was the edge radiation also previously excluded in the FW heat load calculations?

process/data_structure/physics_variables.py
```diff
@@ -1108,6 +1078,9 @@ class PhysicsData:
     pden_plasma_core_rad_mw: float = 0.0
     """total core radiation power per volume (MW/m3)"""
 
+    pden_plasma_core_rad_tauE_mw: float = 0.0
+    """reduced total core radiation power for tauE calculation (MW/m3)"""
```
chris-ashe, line 1082:
```suggestion
    """reduced total core radiation power for τₑ calculation [MW/m³]"""
```

process/models/costs/costs.py
```diff
@@ -2854,7 +2854,8 @@ def coelc(self):
 
         #  Capital recovery factor
 
-        crfcdr = (fefcdr * self.data.costs.discount_rate) / (fefcdr - 1.0e0)
+        # crfcdr = (fefcdr * self.data.costs.discount_rate) / (fefcdr - 1.0e0)
+        crfcdr = 1.0
```
chris-ashe, line 2858: Not sure why this is now just set to 1

process/models/physics/confinement_time.py
```diff
@@ -166,8 +167,16 @@ def calculate_confinement_time(
         try:
             model = ConfinementRadiationLossModel(int(self.data.physics.i_rad_loss))
 
+            # pden_plasma_core_rad_mw was reduced here, but if full radiation, not used!
+            # rad_reduction_for_tauE_only option allows full radiation model in PPB,
+            # but reduced radiation here in confinement time calculation
             if model == ConfinementRadiationLossModel.FULL_RADIATION:
-                p_plasma_loss_mw -= self.data.physics.pden_plasma_rad_mw * vol_plasma
+                if impurity_radiation.rad_reduction_for_tauE_only:
```
chris-ashe, line 174: Should this not be in the `ConfinementRadiationLossModel.CORE_ONLY` condition below?

process/models/physics/impurity_radiation.py
```diff
@@ -734,16 +817,37 @@ def integrate_radiation_loss_profiles(self):
...
         # and github.com/ukaea/PROCESS/issues/3968#issuecomment-4935567006
+
+        # Old case: rho multiplication already performed incorrectly in power
+        # density calculation: don't multiply again here
+        # New case (rho_fix): multiply correct power density here by rho
+        # for integration
+        rho = 1.0
+        if rho_fix:
+            rho = self.plasma_profile.neprofile.profile_x
         self.pden_impurity_rad_total_mw = 2.0e-6 * integrate.simpson(
```
chris-ashe, line 828: I believe it is better here to rename these variables to have `_vol_avg` in them as thats the returned value from the integration when multiplying by `rho` at each point. These values are then multiplied by `vol_plasma` to get the total radiation power. This should make it more clear as to what is happening

process/core/caller.py
```diff
@@ -206,9 +206,28 @@ def call_models_and_write_output(self, xc: np.ndarray, ifail: int):
                     OutputFileManager.close_idempotence_files(
                         self.data.globals.output_prefix
                     )
+
+                    # Now idempotent, return
+                    # Pass model caller and opt params for stability constraint evaluation
```
chris-ashe, line 211: Not sure about this change, probs best if someone like @timothy-nunn checks it

process/data_structure/numerics.py
```diff
@@ -42,20 +42,22 @@ class SolverOutputCondition(IntEnum):
...
     """
 
     EVALUATION = (-2, "Evaluation mode (no optimisation)")
     """In this mode, the code will not perform any optimisation, and will instead
     simply evaluate the constraints for the given input parameters, which is useful
     for testing and for evaluating the performance of a given design point without
-    trying to optimise it. Internally, PROCESS uses `fsolve` (a Newton-Krylov/hybrd
+    trying to optimise it.
+    """
+    SOLUTION = (-1, "Solution mode (no optimisation)")
```
chris-ashe, line 54: Did we accidentaly remove solution mode when setting up the `enum` @jonmaddock @timothy-nunn ?

process/models/physics/impurity_radiation.py
```diff
@@ -705,25 +719,94 @@ def calculate_radiation_loss_profiles(self):
...
+        # plasma power balance
+        f_p_plasma_core_rad_reduction_tauE = (
+            self.data.impurity_radiation.f_p_plasma_core_rad_reduction
         )
+        if rad_reduction_for_tauE_only:
+            # Only reduce radiation for tauE calculation
+            f_p_plasma_core_rad_reduction = 1.0
+        else:
+            # Reduce radiation in PPB as well
+            f_p_plasma_core_rad_reduction = f_p_plasma_core_rad_reduction_tauE
```
chris-ashe, line 741: I assume this will be removed once we put the PR up for the PPB fix?

process/models/physics/impurity_radiation.py
```diff
@@ -705,25 +719,94 @@ def calculate_radiation_loss_profiles(self):
...
+                self.plasma_profile.neprofile.profile_x
+                >= self.data.impurity_radiation.radius_plasma_core_norm
+            )
+            fradedge_profile[edge_mask] = 1.0  # Edge region gets full value
+            pden_impurity_rad_edge_total = (
+                self.pden_impurity_radiation_profile * fradedge_profile
+            )
+
+            # Total radiation profile (core + edge)
+            pden_impurity_rad_total = (
```
chris-ashe, line 774: If you multiply the core and edge `pden` by the total plasma volume and then add them together does it equal `pden_impurity_rad_total * vol_plasma`?

process/models/physics/impurity_radiation.py
```diff
@@ -734,16 +817,37 @@ def integrate_radiation_loss_profiles(self):
...
+        rho = 1.0
+        if rho_fix:
+            rho = self.plasma_profile.neprofile.profile_x
         self.pden_impurity_rad_total_mw = 2.0e-6 * integrate.simpson(
-            self.pden_impurity_rad_profile,
+            self.pden_impurity_rad_profile * rho,
             x=self.plasma_profile.neprofile.profile_x,
             dx=self.plasma_profile.neprofile.profile_dx,
         )
         self.pden_impurity_core_rad_total_mw = 2.0e-6 * integrate.simpson(
```
chris-ashe, line 833: For the core value I think we need to be explicit in stating that this is till done as a full plasma volume integral with the mask of 0's where the core would be. This has always caused confusion as to if the core or edge value is only done as a function of its own volume or that of the total plasma

process/main.py
```diff
@@ -446,18 +448,25 @@ def run_scan(self):
...
-        # i_process_run_mode == 1: optimisation
-        if self.data.numerics.i_process_run_mode == PROCESSRunMode.OPTIMISATION:
+        # ioptimz == 1: optimisation
+        if self.data.numerics.ioptimz == PROCESSRunMode.OPTIMISATION:
             pass
-        # i_process_run_mode == -2: evaluation
-        elif self.data.numerics.i_process_run_mode == PROCESSRunMode.EVALUATION:
-            # No optimisation:
-            # solve equality (consistency) constraints only using fsolve (HYBRD)
+        # ioptimz == -1: solution
```
chris-ashe, line 454: Again will leave this one for @timothy-nunn

chris-ashe (changes requested):

process/core/caller.py
timothy-nunn: This PR does not include objective function 20 so these changes are unnecessary at this point.

Also, `call_models_and_write_output` returns the objective value and normalised residuals, but I cannot see them being used anywhere in this PR. 

I suggest all of these changes in caller can be left for a later PR.

process/data_structure/numerics.py
timothy-nunn: Solution mode was removed a long time ago, it was replaced by evaluation mode. This is because solution mode did not produce consistent solutions and so was essentially useless (except for testing, see https://github.com/ukaea/PROCESS/discussions/4044)

process/models/costs/costs.py
jonmaddock: This calculation would often cause div by 0 errors.

process/data_structure/physics_variables.py
jonmaddock: Happy to do in the actual PR for this.

process/models/physics/confinement_time.py
jonmaddock: My argument here is that we should always use the full radiation in the plasma power balance, and the optional reduction only in the tau E calculation. You are right to point this out, but this is only done here to show the cumulative effect of these various changes when plotting.

process/models/physics/impurity_radiation.py
jonmaddock: I agree.

process/core/caller.py
jonmaddock: Sorry, this is an artefact of other changes on my fork and shouldn't have been in this PR. They can be ignored here.

process/data_structure/numerics.py
jonmaddock: On my fork I have optimisation, solution (solve equalities) and evaluation (just run models until idempotent). I think it's worth adding them back in, but this is for another PR.

process/models/physics/impurity_radiation.py
jonmaddock: Yes.

process/models/physics/impurity_radiation.py
jonmaddock: I think this is shown in the first plot.

process/models/physics/impurity_radiation.py
jonmaddock: I agree, good point.

process/main.py
jonmaddock: My mistake: irrelevant for this PR.

PR #4552, opened 2026-08-25
Updated feasible_point to search over scan points

### LINKED ISSUE #4550: `create.py` only lets you create a new IN.DAT from the first point in a scan
After running a scan with 31 points, I couldn't create a new IN.DAT from an input file other than the very first point, even with using the `-fpi` argument.

When looking at `create.py` I believe it's due to the `feasible_point` method scanning over mfile values rather than scan points.

j-a-foster:
## Description

Updates `create.py` to allow creating new IN.DATs from any point in a scan.

## Checklist

### DIFF
--- a/process/core/io/in_dat/create.py
+++ b/process/core/io/in_dat/create.py
@@ -47,14 +47,12 @@ def feasible_point(filename, position: int):
 
     check_point = 1
 
-    for value in mfile_data.data:
-        # Look for feasible scan points (with ifail = 1)
-        if "ifail" in value and "vmcon_error_flag_(ifail)" not in value:
-            if mfile_data.get(value, scan=check_point) == 1:
-                scan_point = check_point
-            if check_point == position:
-                break
-            check_point += 1
+    for scan in range(1, num_scans + 1):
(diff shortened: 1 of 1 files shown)


process/core/io/in_dat/create.py
timothy-nunn, line 1: So I actually think this entire function could do with a re-work. 

Even with your fixes there are still a number of issues:
- Negative indexing only works for -1, any other index defaults to scan point 1. 
- Scan point 0 silently updates to 1.
- Scan point > number of scans will select the last scan. 

I'm also not sure I like the for loop looking for a scan. I think this logic should be simplified so that:
1. We only support scans `[1, num_scans]` and `-1`, anything else raises an error.
2. We only check the requested scan (`position`) for feasibility and error if it is infeasible. In my opinion, if the user wants a scan point they should explicitly index it, we shouldn't be silently searching for a feasible one.

Let me know if you disagree, have any suggestions, or questions.

timothy-nunn (changes requested):

timothy-nunn (approved):
Thanks Jack, this all works well

PR #4555, opened 2026-08-25
Move data structure to its own file

timothy-nunn:
This is part of the `Parameter`ification of PROCESS however this change is very prone to causing conflicts whenever anyone changes anything. It can be moved into PROCESS main for my own sanity without the `Parameter` being ready yet.


je-cook (commented):
The split out itself I agree with but is it not more appropriate for this to go into the data structure folder as eg base.py or something?

je-cook (approved):

PR #4559, opened 2026-08-27
Write tracker CI job as hatch script

timothy-nunn:


.github/workflows/process.yml
clmould: Can this be hatch-ified too please?

clmould (changes requested):

clmould (approved):

PR #4560, opened 2026-08-27
Support a local repository of regression test assets

timothy-nunn:


tests/regression/test_process_input_files.py
```diff
@@ -327,9 +328,25 @@ def test_input_file(
...
+        )
+    else:
+        # Note that the MFILE must be present in the local repository with the exact
+        # naming pattern as the IN.DAT (including capitalisation, IN.DAT and MFILE.DAT
+        # must be capitals):
+        # - MY.IN.DAT -> MY.MFILE.DAT
+        # - MY_IN.DAT -> MY_MFILE.DAT
+        # - myIN.DAT -> myMFILE.DAT
+        # This naming convention is consistent with PROCESS but is inconsistent with the
+        # remote repoistory for legacy reasons (i.e. scenario_name does not follow
```
clmould:
```suggestion
        # remote repository for legacy reasons (i.e. scenario_name does not follow
```

tests/regression/test_process_input_files.py
clmould: Can you update this to include the `--local-repository` option plz

clmould:
```suggestion
    :param opt_params_only: if True, user specified that only optimisation parameters
```

sorry i just saw that

clmould (changes requested):
v minor , but happy otherwise :)

clmould (approved):

PR #4567, opened 2026-08-28
Change set_filenames in single_run fixture to use file name instead of folder path

clmould:
## Description

Previously this test was using the `tmp_path` in `set_filenames`, leading to files being created with the name of the test inside a folder of the same name:
image

This change means the files created during a run of pytest now have the name of the file preserved:
image

## Checklist

### DIFF
--- a/tests/unit/test_main.py
+++ b/tests/unit/test_main.py
@@ -36,7 +36,7 @@ def single_run(monkeypatch, input_file, tmp_path):
     single_run.filepath = tmp_path
     single_run.models = None
     single_run.data = DataStructure()
-    single_run.set_filenames(tmp_path)
+    single_run.set_filenames(temp_input_file)
     single_run.initialise()
     return single_run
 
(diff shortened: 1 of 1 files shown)


tests/unit/test_main.py
```diff
@@ -36,7 +36,7 @@ def single_run(monkeypatch, input_file, tmp_path):
     single_run.filepath = tmp_path
     single_run.models = None
     single_run.data = DataStructure()
-    single_run.set_filenames(tmp_path)
+    single_run.set_filenames(temp_input_file)
```
timothy-nunn:
```suggestion
    single_run.set_filenames(None)
```

In normal PROCESS runs I see that the argument to  `set_filenames` is `None` because it is actually the custom MFILE path. Fixes to #4011 could cause issues if we do not pass in an MFILE path.

timothy-nunn (changes requested):

timothy-nunn (approved):

PR #4592, opened 2026-09-10
Refactor fusion reactions for dataclass

chris-ashe:
## Checklist


tests/unit/models/physics/test_fusion_reactions.py
```diff
@@ -255,3 +257,121 @@ def test_beam_reaction_rate_coefficient():
     )
 
     assert beam_reaction_rate == pytest.approx(7.465047902975452e-22)
+
+
+def _make_beam_fusion_reactions():
+    return reactions.BeamReactions(data=DataStructure())
```
timothy-nunn: This should be a fixture rather than a helper function

process/models/physics/fusion_reactions.py
```diff
@@ -63,7 +65,45 @@
...
+    """
+
+    pden_plasma_alpha_vol_avg_mw: float = 0.0
+    pden_non_alpha_charged_vol_avg_mw: float = 0.0
+    pden_neutron_vol_avg_mw: float = 0.0
+    fusden_vol_avg: float = 0.0
+    fusden_plasma_alpha_vol_avg: float = 0.0
+    fusden_plasma_protons_vol_avg: float = 0.0
+
+    def __iadd__(self, other: "FusionYieldDensities") -> Self:
```
timothy-nunn: I would add some validation to this method which raises an error if `other` is not an instance of `type(self)`

timothy-nunn: Can we also add a specific unit test just dedicated to testing the addition of two `FusionYieldDensities` works correctly

timothy-nunn:
```suggestion
    def __iadd__(self, other: FusionYieldDensities) -> Self:
```

Add this line to the top of the file also
```python
from __future__ import annotations
```

process/models/physics/fusion_reactions.py
```diff
@@ -63,7 +65,45 @@
...
+        self.pden_plasma_alpha_vol_avg_mw += other.pden_plasma_alpha_vol_avg_mw
+        self.pden_non_alpha_charged_vol_avg_mw += other.pden_non_alpha_charged_vol_avg_mw
+        self.pden_neutron_vol_avg_mw += other.pden_neutron_vol_avg_mw
+        self.fusden_vol_avg += other.fusden_vol_avg
+        self.fusden_plasma_alpha_vol_avg += other.fusden_plasma_alpha_vol_avg
+        self.fusden_plasma_protons_vol_avg += other.fusden_plasma_protons_vol_avg
+        return self
+
+
+class PlasmaReactions:
```
timothy-nunn: If this is a model in `main.py` it should inherit from `process.core.model.Model` like other models. 

Take a look at another model but this means it should not explicitly be passed the data structure.

process/models/physics/fusion_reactions.py
```diff
@@ -584,18 +551,431 @@ def set_physics_variables(self):
...
+            self.outfile,
+            "Volume-averaged proton rate density [particles/m³/sec]",
+            "(fusden_plasma_protons_vol_avg)",
+            self.data.physics.fusden_plasma_protons_vol_avg,
+            "OP ",
+        )
+        po.oblnkl(self.outfile)
+
+
+class BeamReactions:
```
timothy-nunn: Same comment here

https://github.com/ukaea/PROCESS/pull/4592/changes#r4045131410

process/models/physics/fusion_reactions.py
timothy-nunn, line 1: We need to create an issue to remove the state in these models. Any `self.xyz` data should be in the data structure!

timothy-nunn (commented):
Will need a modeller to check some of the model changes (even if results have not changed) and new tests

Regression test failing:
```
FAILED tests/regression/test_process_input_files.py::test_input_file[stellarator_helias] - AttributeError: module 'process.models.physics.fusion_reactions' has no attribute 'FusionReactionRate'
```

Some unit tests are also failing

timothy-nunn (changes requested):

mkovari:
> Will need a modeller to check some of the model changes (even if results have not changed) and new tests

If this is purely a refactor there shouldn't be any model changes.  Can you clarify?

